PR #18, opened 2025-10-02
fix: overall compilation and structure update

Clemensito:
Update versions for SDK, AGP, Kotlin, Java, Jacoco and the Kotlin JVM. The project is now compatible with Android 15. Also implement SonarCloud correctly with project to implement it into the CI.


cocoStein:
lgtm

PR #29, opened 2025-10-02
feat(navigation): set up navigation framework

j-meric:
Centralize routes and navigation actions using a root NavHost and a sealed Screen type.


Clemensito:
I recommend we call it  "Authentication"  rather than"Sign In", what do you think ? Except for that everything looks good.

Clemensito (approved):

PR #30, opened 2025-10-02
feat(model):add AccountService interface for auth management

j-meric:
This interface serves as the basis for the implementation of google authentication.


Clemensito:
Looking like a good base for our authentication process. We may need to follow-up with new methods but for the moment it LGTM

Clemensito (approved):
Looking like a good base for our authentication process. We may need to follow-up with new methods but for the moment it LGTM

PR #43, opened 2025-10-04
feat(testing): Add FirebaseEmulator class

stefantaga24:
Copied the firebase emulator class and its dependencies such that we do not have to handle the connections for emulator tests.

The packagingOptions modification is required, otherwise no tests will run. This is taken from the bootcamp code as well.


stefantaga24:
Yes it seems like so, I will have to see what is wrong with main in order to merge this request!

PR #46, opened 2025-10-04
feat(items): add Item model and ItemsRepository interface

### LINKED ISSUE #17: Implement db functions to store inputs
**Backend Tasks :**

- Design/update database to store clothes details (clothing item, brand, link, etc.).

**Testing & QA :**

-  Unit tests for backend API (create clothing items).

asunluer:
### Description 
This PR is the first step in implementing the Add Items feature.
It focuses on setting up the **data layer** required to store and manage item information in Firestore.

Specifically, it introduces : 
- `Item` data class - represents a clothing item and its details (ID, image, category, type, brand, price, material, link). 
- `ItemsRepository` interface - defines the contract for data operations on items. 

These components lay the foundation for Firestore integration and the upcoming `AddItemsScreen` logic. 

### Changes 
- Added `Items.kt`
- Added `ItemsRepository.kt`

### Testing 
- Successfully built the project using `./gradlew build`
- Verified that new files compile correctly 
- Formatting validated with `./gradlew ktfmtCheck`

### Related issues

Closes #17

### DIFF
--- a/app/src/main/java/com/android/ootd/model/Item.kt
+++ b/app/src/main/java/com/android/ootd/model/Item.kt
@@ -1,0 +1,14 @@
+package com.android.ootd.model
+
+import android.net.Uri
+
+data class Item(
+    val itemId: String,
+    val image : Uri,
+    val category: String,
+    val type: String,
+    val brand : String,
+    val price: Double,
+    val material: String,
+    val link: String,
+)
--- a/app/src/main/java/com/android/ootd/model/Item.kt
+++ b/app/src/main/java/com/android/ootd/model/Item.kt
@@ -4,10 +4,10 @@ import android.net.Uri
 
 data class Item(
     val itemId: String,
-    val image : Uri,
+    val image: Uri,
     val category: String,
     val type: String,
-    val brand : String,
+    val brand: String,
     val price: Double,
     val material: String,
     val link: String,
--- a/app/src/main/java/com/android/ootd/model/Item.kt
+++ b/app/src/main/java/com/android/ootd/model/Item.kt
@@ -3,12 +3,14 @@ package com.android.ootd.model
 import android.net.Uri
 
 data class Item(
-    val itemUUID: String,
+    val uuid: String,
     val image: Uri,
     val category: String,
     val type: String,
     val brand: String,
     val price: Double,
-    val material: String,
+    val material: List<Material>,
     val link: String,
 )
+
+data class Material(val name: String = "", val percentage: Double = 0.0)
--- a/app/src/main/java/com/android/ootd/model/Item.kt
+++ b/app/src/main/java/com/android/ootd/model/Item.kt
@@ -3,7 +3,7 @@ package com.android.ootd.model
 import android.net.Uri
 
 data class Item(
-    val itemId: String,
+    val itemUUID: String,
     val image: Uri,
     val category: String,
     val type: String,
--- a/app/src/main/java/com/android/ootd/model/ItemsRepository.kt
+++ b/app/src/main/java/com/android/ootd/model/ItemsRepository.kt
@@ -1,50 +1,48 @@
 package com.android.ootd.model
 
-/**
- *  Repository class that handles data operations for items.
- */
+/** Repository class that handles data operations for items. */
 interface ItemsRepository {
 
-    /** Generates and returns a new unique identifier for a Clothing item. */
-    fun getNewItemId(): String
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/android/ootd/model/ItemsRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+   */
+  suspend fun getAllItems(): List<Item>
+
+  /**
+   * Gets a specific item by its unique identifier.
+   *
+   * @param id The unique identifier of the item to retrieve.
+   * @return The item with the specified identifier, or null if not found.
+   */
+  suspend fun getItemById(id: Long): Item?
```
cocoStein, line 22: What was your thought process about making the output nullable ? Isn't it easier to handle the case where we try to access a non existing item in the fun and propagate an error ? 

And also the `uuid` of the items is a `String`. 
I like the name of the fun it's very explicit

app/src/main/java/com/android/ootd/model/Item.kt
```diff
@@ -0,0 +1,14 @@
+package com.android.ootd.model
+
+import android.net.Uri
+
+data class Item(
```
cocoStein, line 5: The overall implementation of the data class looks good. A few thoughts:

- Should we add a val ownerID: String parameter to help reference items for future interactions?
- Consider renaming itemId to itemUUID to make it explicit that each item is unique.
- For the material field, some clothes are made of multiple materials. Maybe we could use a dictionary/map like  (material name) -> (% composition)?


Additionally, we might want to create custom types to ensure validity (e.g., making sure link starts with http://). Let's discuss these ideas with the team.

cocoStein (changes requested):
Except theses small changes the PR looks good, nice first PR ;)

app/src/main/java/com/android/ootd/model/ItemsRepository.kt
asunluer: Thank you for your review! You are absolutely right - the `id` should be a `String` :). 

Regarding the nullable return type, I initially thought of handling the `null` case when calling the function but I agree that throwing an error when the item is not found would be a better approach.

app/src/main/java/com/android/ootd/model/Item.kt
asunluer: Thank you so much. So I have a few questions and thoughts based on your feedback : 

- I am not sure I understand the need for `ownerId`. So, could you explain a bit more how it would be used or what kind of interactions you are thinking of?   
- I'll rename the `itemId` to `itemUUID`. Could you tell me their difference ? 
- About the `materiel` field: maybe I wasn't clear enough  during our Friday afternoon meeting, but this is what I was referring. So, we can do a list of `Map<Int, String>`.  Do you think this might overwhelm the user while filling the fields? 

Regarding the validity of link, I was thinking of implementing a function called `isValidLink` in  `AddItemViewModel`, since, that's where the user interaction happens and we need to update the state of the link field. 

We can definitely talk with the team also :).

app/src/main/java/com/android/ootd/model/ItemsRepository.kt
cocoStein: For me either ways are fine, we just have to be consistent around the project 👍🏼

app/src/main/java/com/android/ootd/model/Item.kt
cocoStein: - On `ownerId`: My intent was to use ownerId to identify who created the item in the database. Alternatively, we could rely on the user’s inventory items. I’m not sure that’s necessary, what do you think?
- On `itemUUID`: There’s no functional difference for me; using `itemUUID` makes it explicit that it’s a unique identifier. But it's just a preference, and it's your PR so you should choose.
- Material: Your proposal is solid. However, do we really want to add a material field? Personally, I don’t know the composition of most of my clothes, and as you mentioned, filling every field thoroughly would be tedious. What would we do with this information ? The only use case is for specific types of material like: `100% real leather` ?
- For the validity, as we briefly discussed, it's a M2 problem, we can do a simple `isValidLink ` as you proposed but i think later on we need to implement our own `Refined Types` [example of refined types in scala](https://github.com/Iltotore/iron)

cocoStein: Yes you're right 👍🏼

PR #52, opened 2025-10-05
Add profile page

stefantaga24:
I implemented the ui for the profile page without the programming logic behind it to make it work. This way we can speed up the review process and not review the ui + logic at the same time.

This is how the profile page will look like:
image

This of course will be a smaller part of the search page and there it will look like this:

image


app/src/main/java/com/android/ootd/ui/search/UserProfileCard.kt
```diff
@@ -0,0 +1,68 @@
...
+import androidx.compose.ui.unit.sp
+import com.android.ootd.model.user.User
+import com.android.ootd.ui.theme.OOTDTheme
+
+@Preview
+@Composable
+fun UserProfileCardPreview() {
+  OOTDTheme {
+    UserProfileCard(
+        selectedUser = User(uid = "Bob", name = "Micheal", friendList = emptyList()),
```
MarcK0909, line 31: small typo: "Micheal" should be "Michael"

MarcK0909 (changes requested):
LGTM, just fix the typo

MarcK0909 (approved):

PR #57, opened 2025-10-06
Feature/authentication

j-meric:
- Change signInWithGoogle parameter from string to credentials to enable unit testing without statically mocking SDK methods.

- Change signOut return type to Result<Unit> so callers can observe and handle sign-out failures instead of assuming success.


Clemensito (approved):
Looks good !

Issue #56, opened 2025-10-06
Add feed screen with upload post button


bbianca2004:
Each user should be able to upload a post once they log into the app

PR #61, opened 2025-10-06
feat(feed): add feed architecture and mock UI

bbianca2004:
Implements the foundational structure for the feed feature using MVVM.

#### Includes:
- `OutfitPost` data model
- `FeedRepository` interface, Firestore implementation, and provider
- `FeedViewModel` and `FeedScreen` with conditional post visibility
- `OutfitPostCard` composable for displaying posts

This setup establishes the feed's data flow and UI logic.
Firebase integration still needs testing, and FeedRepository function design 
to be agreed on with the team.

Related to #56

### DIFF
--- a/app/src/androidTest/java/com/android/ootd/model/feed/FeedRepoFirestoreTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/model/feed/FeedRepoFirestoreTest.kt
@@ -95,7 +95,14 @@ class FeedRepoFirestoreTest {
   }
 
   @Test
-  fun hasPostedToday_defaultFalse() = runBlocking { assertEquals(false, repo.hasPostedToday()) }
+  fun hasPostedToday_returnsTrueWhenUserHasPostedToday() = runBlocking {
+    val userId = "user-today"
+    val post = samplePost("today-post", ts = System.currentTimeMillis()).copy(uid = userId)
+    repo.addPost(post)
+
+    val result = repo.hasPostedToday(userId)
+    assertTrue(result)
+  }
 
   @Test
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/ootd/model/OutfitPost.kt
```diff
@@ -0,0 +1,26 @@
...
+ * @property timestamp time passed since posting
+ */
+data class OutfitPost(
+    val postUUID: String,
+    val userName: String,
+    val userID: String,
+    val userProfilePicURL: String, // firebase url of the user's profile picture
+    val outfitURL: String,
+    val description: String,
+    val itemsID: List<Item>,
```
MarcK0909, line 22: The name itemsID should be changed to just "items" if it stores items or if we store the ID shouldn't the type of the list be String ?

app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+      if (_hasPostedToday.value) {
+        _feedPosts.value = repository.getFeed()
+      }
+    }
+  }
+
+  fun onPostUploaded() {
+    viewModelScope.launch {
+      _hasPostedToday.value = true
+      _feedPosts.value = repository.getFeed()
```
MarcK0909, line 36: Feed only reloads once when user uploads but we should change this so that it reloads more often to allow the user to view posts after he posts

MarcK0909 (changes requested):
Overall good work just some little tweaks for the Post data class to resolve

PR #62, opened 2025-10-06
Feature/UI theme

j-meric:
**Add typography tests, fix Bodoni implementation, and add Figma color palette**

- Added unit tests for Bodoni typography (displayLarge, titleLarge, bodySmall). 

- Fixed the Bodoni FontFamily implementation to use correct weights so typography renders as intended. 

-  Added the app color palette from the Figma design (Primary, Secondary, Tertiary, Last, Background).


MarcK0909 (approved):
LGTM and good idea to have implemented the tests. Nice job !

PR #65, opened 2025-10-07
Feed filtering

MarcK0909:
**Summary**
Added filtering to the main feed so that the user only sees posts made by people in his friend list. Also modified the OutfitPost class to have default values (might be changed later).

**Tests**
Added both simple unit tests to test the filtering as well as firestore ones to get a better idea if everything is working end-to-end.


app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
```diff
@@ -25,15 +33,44 @@ class FeedViewModel : ViewModel() {
...
+    val user = _currentUser.value
+    if (user == null) {
+      _feedPosts.value = emptyList()
+      return
+    }
+    _feedPosts.value = filterPostsByFriends(_allPosts.value, user)
+  }
+
+  // Filters to include only posts from friends
+  fun filterPostsByFriends(posts: List<OutfitPost>, user: User): List<OutfitPost> {
```
bbianca2004, line 70: This function does not depend on any view model state, might be worth it moving it to some utils file later on for reusability. But that is for future reference, at the moment it is good here!

PR #106, opened 2025-10-13
Feat: Navigation upon sign-In

Clemensito:
# Summary
Implemented the navigation that permits the user to get directed to the "Register" page if he does not have a proper profile (username).

# What
The code sends a request to the database that asserts if yes or no the user's ID has a username attached to it. If not he is redirected to register, else he is sent to the "Overview" page where his feed should be

- ViewModel
  - Creation of 'userExists'  method inside UserRepository that asserts if the user is registered or not
  - Creation of a new method 'routeAfterGoogleSignIn' that dictates the screen that the user will be sent to upon clicking sign-in button
- Documentation
  - Kdoc comments added for 'userExists' and 'routeAfterGoogleSignIn'
- Tests
  - Integration tests for the whole implementation, no unit nor end to end

# Why
It is a necessary feature for the app

# Checklist

### DIFF
--- a/app/src/main/java/com/android/ootd/MainActivity.kt
+++ b/app/src/main/java/com/android/ootd/MainActivity.kt
@@ -84,7 +84,7 @@ fun OOTDApp(
       composable(Screen.Authentication.route) {
         SignInScreen(
             credentialManager = credentialManager,
-            onSignedIn = { navigationActions.navigateTo(Screen.Overview) },
+            onSignedIn = { navigationActions.navigateTo(Screen.Feed) },
             onRegister = { navigationActions.navigateTo(Screen.RegisterUsername) })
       }
     }


app/src/main/java/com/android/ootd/ui/authentication/SignInScreen.kt
```diff
@@ -52,6 +52,7 @@ fun SignInScreen(
     authViewModel: SignInViewModel = viewModel(),
     credentialManager: CredentialManager = CredentialManager.create(LocalContext.current),
     onSignedIn: () -> Unit = {},
+    onRegister: () -> Unit = {}
```
j-meric: **Nitpick**: Document the parameters onSignedIn and onRegister so it is easier to differentiate them. 

```
/**
 * A composable function for the SignIn screen, which includes functionality for user login.
 *
 * @param authViewModel The ViewModel that manages authentication state. Defaults to a view model instance for SignIn.
 * @param credentialManager Manages user credentials, created from the current context.
 * @param onSignedIn Callback invoked when the user successfully signs in to the app.
 * @param onRegister Callback invoked to navigate to the registration screen.
 */
```

PR #129, opened 2025-10-15
Fix: Date formatting, and color for datePicker

Clemensito:
# Summary
Changed the datePicker field to be more user friendly and date is now EU format

# What
format the date to be EU standard and the picker uses now bodoni. As well as making the datepicker textField clickable

- UI
 - Updated colors for Dismiss and Confirm buttons, as well as font bodini.

# Why
Much needed change for UX

# Checklist


cocoStein (approved):
Good job finding this bug 🔥 

Keep on 

Image

PR #142, opened 2025-10-16
fix(proguard): Fix proguard rules because they obfuscate firebase dat…

stefantaga24:
Due to the fact that proguard is obfuscating code, the name of the dataclasses are also being obfuscated and the implicit serialization for firebase does not work anymore. This would make the search screen fail when trying to access the user repository.

The exact error is the following:

10-17 01:39:11.258 15104 15104 E UserRepositoryFirestore: Error transforming document v8dYB5iAEaew2mHCzYKHN8aqPyd2: Could not deserialize object. Class h6.c does not define a no-argument constructor. If you are using ProGuard, make sure these constructors are not stripped
10-17 01:39:11.258 15104 15104 E UserRepositoryFirestore: java.lang.RuntimeException: Could not deserialize object. Class h6.c does not define a no-argument constructor. If you are using ProGuard, make sure these constructors are not stripped

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -19,3 +19,20 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+# Keep all classes annotated with @Keep
+-keep @androidx.annotation.Keep class * { *; }
+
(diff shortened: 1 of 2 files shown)


a1exxd0:
lgtm

Issue #126, opened 2025-10-15
Refactor test classes to use same base class


stefantaga24:
We had some problems that authentication tests were closing the firebase when that should not be happening. As well, different tests use different repositories and open/close the firebase when they see fit. This might lead to problems when pushing to the CI, therefore the design needs to be united for better development.

PR #151, opened 2025-10-24
refactor(tests): refactor user tests to be uniformized

stefantaga24:
# Summary
The user repository in tests will be named userRepository to distinguish it from others. Also renamed RegisterUserRepository to RegisterUserRepositoryTest.

# What
In the test class FirestoreTest, now I changed the name of the user repository to userRepository instead of what it was before which was the generic repository.

- Tests
  - Tests for user repository were updated in order to comply to general guidelines for tests.

# Why
This is related to issue https://github.com/swent-Team01/OOTD/issues/126 and it helps with removing code duplication and with the building of future tests. New people that want to create tests will just use the repositories that are already initialized in BaseTest by inheriting the FirestoreTest class.

For the guidelines check this link: [Testing guidelines](https://github.com/swent-Team01/OOTD/wiki/Testing-using-repositories-guidelines)


Clemensito (approved):
LGTM 🔥

PR #153, opened 2025-10-24
refactor(tests): refractor tests feed repository

stefantaga24:
# Summary
Changed the tests for the feed repository to follow the testing guidelines.

# What
The class FeedRepositoryFireTest, FirestoreTest and BaseTest were modified in order for the feed repository to be initialised in only one place.

- Tests
  - Feed repository tests were modified to use the specific feed repository now initialized in BaseTest.

# Why
This is related to issue https://github.com/swent-Team01/OOTD/issues/126 and it helps with removing code duplication and with the building of future tests. New people that want to create tests will just use the repositories that are already initialized in BaseTest by inheriting the FirestoreTest class.

For the guidelines check this link: [Testing guidelines](https://github.com/swent-Team01/OOTD/wiki/Testing-using-repositories-guidelines)


MarcK0909 (approved):
Looks very good to me, tests are more clear now and the new base class will make it easier to write the next tests. Very solid work and thx

PR #158, opened 2025-10-25
feat(EditItemsScreen): implement dynamic image resizing and improve U…

cocoStein:
# Summary
Added new Responsive scrolling to the add and edit item screen so that user with smaller screens can better see the picture of the current item while editing it.

This is inspired from the google official documentation: https://developer.android.com/develop/ui/compose/touch-input/pointer-input/scroll#result

Added a new `rememberImageResizeScrollConnection` function to handle the nested scroll with a top image
# What
Now the image can always been seen while editing the different text fields of the add/edit item. This is in response to some User flow testing i did and noticed some people going back and forth to see the picture and the details in it.

[Screen_recording_20251025_171033.webm](https://github.com/user-attachments/assets/a3578e00-1d97-4bf9-b0d2-e87b51d54fa8)
[Screen_recording_20251025_171052.webm](https://github.com/user-attachments/assets/f047e877-9c07-4e59-bb28-983fe2138234)

- Tests
  - Modified the tests so they can properly handle the new UI
  - For the `clickingAddItemReturns` test i had to recreate the `enterAddItemDetails` fun because it didn't set properly the new fields, but as we have the new tests structure this should not create any problem

# Why
So that people with smaller phones can still see the pictures of the item while editing the fields.

# How to test
1. Use the `smallEmulator` on android Studio
2. `Edit` an existing Item on a post or `Add` a new Item, and scroll

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/post/AddItemsScreen.kt
+++ b/app/src/main/java/com/android/ootd/ui/post/AddItemsScreen.kt
@@ -48,14 +48,10 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.focus.onFocusChanged
-import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.graphicsLayer
-import androidx.compose.ui.input.nestedscroll.NestedScrollConnection
-import androidx.compose.ui.input.nestedscroll.NestedScrollSource
 import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalContext
-import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.LocalFocusManager
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.painterResource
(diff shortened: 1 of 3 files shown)


app/src/test/java/com/android/ootd/ui/post/EditItemsScreenTest.kt
```diff
@@ -289,8 +289,9 @@ class EditItemsScreenTest {
   fun `buttons have proper text labels`() {
     composeTestRule.setContent { EditItemsScreen(editItemsViewModel = mockViewModel) }
 
-    composeTestRule.onNodeWithText("Select from Gallery").assertIsDisplayed()
-    composeTestRule.onNodeWithText("Take a new picture").assertIsDisplayed()
-    composeTestRule.onNodeWithText("Save Changes").assertIsDisplayed()
+    composeTestRule.onNodeWithText("Select from Gallery").assertExists()
```
MarcK0909: Why use assertExists instead of IsDisplayed, this is a less thorough test unless you have a specific reason. for doing so otherwise I think it's best to revert to check if displayed

cocoStein: As they are now in a `Lazycolumn` they may not be displayed at all times. If you use `assertIsDisplayed()` you need to add extra composeTestRule in order to scroll to the element and to click on the right button to obtain the elements. Which in my point of view needs to be checked in the Connected tests (which they are) , the UnitTests just look if the elements exist.

Tell me what you think of it 👍🏼

MarcK0909: Ok then I agree it seems to be enough since the connected test do it more thoroughly

PR #159, opened 2025-10-25
feat: refactor rules

stefantaga24:
# Summary
Refactored code and firebase security rules in order to enforce data protection as detailed in the [Wiki page](https://github.com/swent-Team01/OOTD/wiki/Security-rules-and-guidelines).

# What

- Firebase
  - Changed the security rules such that not all data is available to every user and we have a better separation of concerns. This is implemented by every piece of data having an ownerId.
  - Removed the getFeedPosts function as it was not used anywhere in the code.
  - The ownerId is defined differently than the userId in the case of the Account repository because actually for testing it would be extremely hard to test anything related to the AccountRepository if we use the userId and not the ownerId. During testing right now, we have the current anonymous user be the ownerId of all of the users in order for it to modify them accordingly.
- Repositories
  - The items, outfit_posts and posts related repositories have been changed in order for all items to have an ownerId and for it to be created from the Firebase.auth.uid as done in M3. 
  - For the Account repository, we now check if a user exists by querying the User collection. This is because the User collection is the single source of truth for all of the users that are in the application and everyone can access it without restrictions. It only contains the userId and username fields, which means there is no leak of personal information. 
  - Also, I have reversed the way we register a user. First I add the account and then I add the user in the database. This is because adding the account checks if there is already a user and if we do it the other way around this check would fail.
  - Changed EditItems to load the ownerId when loading the item.
- Documentation
  - Added an external wiki page where I detail how the security rules are implemented.
- Tests
  - All firebase emulator tests related to the items, posts and outfit posts have been changed such that they respect the firebase security rules.

# Why
With this change, everyone will be able to work on a general format when needing to add data. Now we will have all of the tests follow the security rules and the team will be able to give access to data only to the users that really need it.

# Manual testing

The testing for this PR has been just running the tests locally and making sure they pass the security rules.

### DIFF
--- a/app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
@@ -10,10 +10,9 @@ import kotlin.collections.get
 import kotlinx.coroutines.tasks.await
 
 const val ITEMS_COLLECTION = "items"
+const val OWNER_ATTRIBUTE_NAME = "ownerId"
 
 class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsRepository {
-
-  private val ownerAttributeName = "ownerId"
 
   override fun getNewItemId(): String {
     return db.collection(ITEMS_COLLECTION).document().id
@@ -24,7 +23,7 @@ class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsReposit
         Firebase.auth.currentUser?.uid
             ?: throw Exception("ToDosRepositoryFirestore: User not logged in.")
     val snapshot =
-        db.collection(ITEMS_COLLECTION).whereEqualTo(ownerAttributeName, ownerId).get().await()
+        db.collection(ITEMS_COLLECTION).whereEqualTo(OWNER_ATTRIBUTE_NAME, ownerId).get().await()
     return snapshot.mapNotNull { mapToItem(it) }
   }
 


app/src/main/java/com/android/ootd/model/feed/FeedRepositoryFirestore.kt
```diff
@@ -13,12 +15,20 @@ import kotlinx.coroutines.withTimeout
...
+            ?: throw Exception("ToDosRepositoryFirestore: User not logged in.")
     return try {
       val snapshot =
           withTimeout(5_000) {
-            db.collection(POSTS_COLLECTION_PATH).orderBy("timestamp").get().await()
+            db.collection(POSTS_COLLECTION_PATH)
+                .whereEqualTo(ownerAttributeName, ownerId)
+                .orderBy("timestamp")
+                .get()
+                .await()
```
bbianca2004: I noticed that FeedRepositoryFirestore.getFeed() now filters by ownerId, returning only posts belonging to the authenticated user.
While this aligns with the new Firebase security rules (which prevent reading other users’ documents), it changes the original intent of this function, which was to fetch all posts for the feed (e.g., a global feed for when we choose to implement this). 
However, I do agree the original scope of the function might outdated as I didn't implement it with security rules in mind, I think that at least the name of the function should be changed to better reflect the new scope( e.g. `getUserPosts()`).

What do you think?

firebase/firestore/firestore.rules
```diff
@@ -38,28 +24,55 @@ service cloud.firestore {
       allow write: if true;
     }
 
+    // Posts: readable by author or mutual friends
+    match /outfit_posts/{postId} {
+      allow read: if isAuthed() &&
+                  (resource == null || resource.data.ownerId == request.auth.uid ||
+                   isFriend(request.auth.uid, resource.data.ownerId));
+
```
bbianca2004: I’m a bit unsure about the use of resource == null in this read condition. 

From what I understand, resource should represent the existing document and is only null during a create operation. So in a read rule, resource == null shouldn’t normally be true, but I’ve seen a few discussion threads mentioning that, in the emulator or some query contexts, resource can unexpectedly evaluate as null even when a document exists.
If that happens, this rule might end up passing the (resource == null || …) part and allow a read even when ownership doesn’t match. There is a slight chance that this misbehaviour could happen, but just to be sure, I would suggest this just to be safe and align with recommended patterns:

`allow read: if isAuthed() && resource.data.ownerId == request.auth.uid;`

Also please let me know if you had a motive for using the resource == null check, what I said previously is just a small concern! You probably know better than me when it comes to the security rules :))

app/src/main/java/com/android/ootd/model/feed/FeedRepositoryFirestore.kt
stefantaga24: Yes, of course, I will make sure to rename it to getUserPosts. Indeed, we are not able to fetch all of the other user's posts because that would not be in accordance with how the app should work, as we want the user to see only friends post and especially only after they have posted.

app/src/androidTest/java/com/android/ootd/model/feed/FeedRepositoryFirestoreTest.kt
```diff
@@ -121,8 +121,8 @@ class FeedRepositoryFirestoreTest : FirestoreTest() {
     // true after posting today by the signed-in user
     val post = samplePost("today-post", ts = System.currentTimeMillis())
     feedRepository.addPost(post)
```
bbianca2004: Could you please tell me why you chose to comment these two lines? The test passes with those uncommented even with the new security rules when I run it

PR #161, opened 2025-10-26
Feature : Added a bottom navigation bar

MarcK0909:
# Summary
This pull request introduces a bottom navigation bar to the app, allowing users to easily switch between the Feed, Search, Inventory, and Profile screens. It includes the implementation of the navigation bar UI, associated navigation logic, and comprehensive tests to ensure correct behavior and visibility. Additionally, an Inventory screen placeholder is added, and the navigation structure is updated to support these changes.

# What
- UI: Added a bottom navigation bar that is visible from the feed, search, profile and the to-be-implemented inventory screen
- Tests: Added both unit and connected tests to ensure the correctness of the code and behavior. The tests check both that components are displayed and that the navigation is handled correctly.

# Why
The addition of this bottom bar ensures a faster and more efficient UX due to limiting the number of clicks he has to perform. For example, one can now simply navigate to the search screen from his profile instead of having to go back to the feed and then to search screen. This was discussed with the PO to ensure the user has a better overall experience and we belive this simple tool achieves that.

# Screenshots / Videos
Screenshot 2025-10-26 at 17 35 03

# Checklist
- [x] Tests: added unit and connected tests
- [x] Manual verification: app run and basic flows checked

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/navigation/BottomNavigationBar.kt
+++ b/app/src/main/java/com/android/ootd/ui/navigation/BottomNavigationBar.kt
@@ -6,7 +6,6 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.outlined.List
 import androidx.compose.material.icons.outlined.Home
-import androidx.compose.material.icons.outlined.List
 import androidx.compose.material.icons.outlined.Person
 import androidx.compose.material.icons.outlined.Search
 import androidx.compose.material3.Icon


app/src/main/java/com/android/ootd/MainActivity.kt
```diff
@@ -63,97 +66,126 @@ class MainActivity : ComponentActivity() {
  * @param credentialManager Default [CredentialManager] instance for authentication flows.
  */
 @Composable
+@Suppress("UnusedMaterial3ScaffoldPaddingParameter")
```
cocoStein: Why have you added this `@Supress` ? 
```suggestion

```
Regarding that on line 102 you do `modifier = Modifier.padding(innerPadding)) {...` the inner padding is being used, so i think you can either remove it or change how you handle the padding for the NavHost
[doc on the @Supress](https://googlesamples.github.io/android-custom-lint-rules/checks/UnusedMaterial3ScaffoldPaddingParameter.md.html)

app/src/test/java/com/android/ootd/ui/navigation/BottomNavigationBarUnitTest.kt
```diff
@@ -0,0 +1,36 @@
+package com.android.ootd.ui.navigation
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class BottomNavigationBarUnitTest {
+
+  @Test
+  fun routeToTab_mapsKnownRoutes_toCorrectTabs() {
```
cocoStein: Very Nice Unit Test, i like them 🔥 easy to comprend and test the core function of the nav bar 🧨

app/src/main/java/com/android/ootd/ui/navigation/BottomNavigationBar.kt
```diff
@@ -0,0 +1,87 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+/** Tabs used by the bottom navigation. Each tab points to a `Screen` destination. */
+sealed class Tab(val name: String, val icon: ImageVector, val destination: Screen) {
+  object Feed : Tab("Feed", Icons.Outlined.Home, Screen.Feed)
+
+  object Search : Tab("Search", Icons.Outlined.Search, Screen.SearchScreen)
+
+  object Inventory : Tab("Inventory", Icons.Outlined.List, Screen.InventoryScreen)
```
cocoStein: Sonar cloud raise an issue here for the `Icons.Outlined.List`, try to have a look at it to see if it can be removed. From what i understand it's just that it should not be used anymore so we should be able to find a work around

cocoStein (changes requested):
Overall great Work 🔥 
This bottom navigation bar significantly improves the ux by reducing 
the number of clicks needed to navigate between main screens. The logic of not showing it on all the screen works well and is cleverly implanted, the Tab sealed class provides a clean, type-safe way to handle navigation. 
The separation between BottomNavigationMenu (UI) and BottomNavigationBar 
(route mapping) is well done.

The combination of unit tests 
(BottomNavigationBarUnitTest) and integration tests (BottomNavigationBarTest) 
provides comprehensive coverage of both the logic and UI behavior.

app/src/main/java/com/android/ootd/ui/navigation/BottomNavigationBar.kt
MarcK0909: I changed to Icons.AutoMirrored.Outlined.List which isn't deprecetated

PR #178, opened 2025-10-29
Fix Edit screen UI and delete button logic

cocoStein:
# Summary
Fixed the UI of the Edit screen to follow the rest of the app style and delete button logic, add the material field

# What
- Delete button logic: added a new state, that when active (once the item is correctly deleted on the firebase) the user is taken back to the previous screen.
- UI: Modified the topbar to follow the rest of the App style and logic, Big "EDIT ITEM" text with the font used in the app and the primary color.
- UI: added the Material field, even if it's not sure the field will stay in a later version it's better to be consistent across the app.
- created a new `BaseItemViewModel` class to parent `AddItemsViewModel` & `EditItemsViewModel`. So they can inherit from this parent class. And avoid code duplication

Added Unit test to test the material field on the edit screen and the logic in the ViewModel.

Modified some tests:
- removed `save callback is triggered when save button is clicked with valid data`: because now the actions it wants to check can't be performed in a unit test, but is already checked in another connected test.
- modified `deleteItem calls repository when itemId is not empty` to follow the new implementation of the del flow
- removed `allUIElementsAreDisplayed`: already done in the Unit test (quicker execution)
- removed `topAppBarIsDisplayedCorrectly`: already done in the Unit test (quicker execution)
- removed `clearingCategoryDisablesSaveButton`: done in the vm test

# Why
So the user has a better flow, he doesn't has to go back by himself. So the edit screen follow the same style as the rest of the app 🔥 

# Screenshots / Videos
Screenshot_20251030_112621

# How to test
- Do a fit check
- Go to the items list
- Click on the pen button to edit an item
- Press the red delete item button 🔥

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/post/AddItemsViewModel.kt
+++ b/app/src/main/java/com/android/ootd/ui/post/AddItemsViewModel.kt
@@ -69,56 +69,58 @@ data class AddItemsUIState(
  */
 open class AddItemsViewModel(
     private val repository: ItemsRepository = ItemsRepositoryProvider.repository,
-) : BaseItemViewModel() {
+) : BaseItemViewModel<AddItemsUIState>() {
 
-  private val _uiState = MutableStateFlow(AddItemsUIState())
-  open val uiState: StateFlow<AddItemsUIState> = _uiState.asStateFlow()
+  override val _uiState = MutableStateFlow(AddItemsUIState())
+  override val uiState: StateFlow<AddItemsUIState> = _uiState.asStateFlow()
   private val _addOnSuccess = MutableStateFlow(false)
   val addOnSuccess: StateFlow<Boolean> = _addOnSuccess
-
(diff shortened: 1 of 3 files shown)


asunluer (approved):
Excellent separation of concerns and architecture.
The use of a generic abstract base ViewModel `BaseItemViewModel<T>` is clean, scalable, and Kotlin-idiomatic.
The subclasses `AddItemsViewModel` and `EditItemsViewModel` correctly implement state-specific logic while reusing shared functionality.
The code is consistent, easy to test, and Compose-friendly thanks to `StateFlow` usage.
This PR is very well-structured and aligns with modern Android + Kotlin best practices.
It’s clear you’ve designed it with code reuse, testability, and clarity in mind.
Thank you so much for your contribution. You did a good job ! 👍🏻

Note: Don't forget to handle the issues that Sonar is mentioning

PR #195, opened 2025-11-01
Fix: UserRepository updates correctly the profilePicture

Clemensito:
# Summary
Quickfix that adds a parameter profilePicture to editUser, and reforactoring the method to be cleaner. As well as the way editUser in the ViewModel is called

# What
AccountViewModel.EditUser now calls both repository methods and userRepository editUser takes profilePicture URL as a parameter. Also changed the error handling to do nothing when the new values are blank

- ViewModel
  - EditUser method to correctly call both updating methods

- Documentation
  - Kdoc comments updated
  
- Tests
  - Unit tests added/updated to match new behavior 

# Why
Needed fix and readability/understandability

# Checklist


app/src/main/java/com/android/ootd/ui/account/AccountViewModel.kt
```diff
@@ -186,22 +186,26 @@ class AccountViewModel(
...
-        if (newUsername.isNotBlank()) userRepository.editUsername(currentUserId, newUsername)
+        accountRepository.editAccount(currentUserId, newUsername, newDate, imageUrl)
+        userRepository.editUser(currentUserId, newUsername, imageUrl)
         // Update UI state with new values
         val updatedState = _uiState.value.copy(isLoading = false, errorMsg = null)
         _uiState.update {
           updatedState.copy(
               username = newUsername.ifBlank { it.username },
+              dateOfBirth = newDate.ifBlank { it.dateOfBirth },
               profilePicture = profilePicture.ifBlank { it.profilePicture })
```
MarcK0909: Shouldn't it be profilePicture = imageUrl.ifBlank { it.profilePicture } as we want to use the uploaded picture and not the local one ?

app/src/main/java/com/android/ootd/ui/account/AccountViewModel.kt
```diff
@@ -247,6 +225,7 @@ class AccountViewModel(
    * @return the download URL of the uploaded image
    */
   private suspend fun uploadImageToStorage(localPath: String, userId: String): String {
+    if (localPath.isBlank()) return localPath
```
MarcK0909: Nitpick: I think it's nicer to return an empty string instead

app/src/main/java/com/android/ootd/model/user/UserRepositoryFirestore.kt
```diff
@@ -142,35 +142,23 @@ class UserRepositoryFirestore(private val db: FirebaseFirestore) : UserRepositor
...
   }
 
-  override suspend fun editUsername(userID: String, newUsername: String) {
+  override suspend fun editUser(userID: String, newUsername: String, profilePicture: String) {
     try {
-      // Validate input
-      if (userID.isBlank() || newUsername.isBlank()) {
-        throw IllegalArgumentException("User ID and username cannot be blank")
-      }
-      // Check if user exists
```
MarcK0909: Is there a reason to remove this check ?

app/src/main/java/com/android/ootd/ui/account/AccountViewModel.kt
Clemensito: bruh the localPath is an empty string in that case

app/src/main/java/com/android/ootd/ui/account/AccountViewModel.kt
```diff
@@ -186,22 +186,26 @@ class AccountViewModel(
...
+   * @param newUsername the new username to set, blank by default
+   * @param newDate the new date to set, blank by default
+   * @param profilePicture the new profile picture to set, blank by default
    */
   fun editUser(newUsername: String = "", newDate: String = "", profilePicture: String = "") {
     viewModelScope.launch {
       _uiState.update { it.copy(isLoading = true) }
 
       try {
         val currentUserId = accountService.currentUserId
```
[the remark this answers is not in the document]
MarcK0909: Imo it wouldn't be a big change so if possible pls do it this sprint to avoid issues we may have.

app/src/main/java/com/android/ootd/model/user/UserRepositoryFirestore.kt
MarcK0909: Firstly they don't throw the same error (IllegalArgumentException as opposed to NoSuchElementException) and also get user does not validate newUsername so at bare minimum keep the check for newUsername

app/src/main/java/com/android/ootd/ui/account/AccountViewModel.kt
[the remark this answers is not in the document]
Clemensito: It is, it makes many tests fail. I'll to look into that

app/src/main/java/com/android/ootd/model/user/UserRepositoryFirestore.kt
MarcK0909: I get your point, but those are still two different layers of validation. Logging “user not found” describes a lookup failure, while input validation ensures we don’t even reach that point with invalid data. The takeIf logic only handles whether we replace the username, not whether the provided one is valid. Keeping a simple isBlank check makes that distinction explicit and prevents sending empty or malformed usernames to Firestore in the first place. It’s a small safeguard, but it helps keep the API behavior predictable and consistent.

PR #206, opened 2025-11-03
feat: add follow notification

stefantaga24:
# Summary
Added follow notification mechanism that goes both ways. When one user A adds user B, a follow notification is sent and if user B accepts, both will have each other in the friend list.

# What

- UI
  - Added the notifications screen where we will be able to see all follow notifications and what other notifications might be implemented in the future.
  - The notification screen does not ask for notification permissions yet because push notifications will be implemented in a new PR. 
- ViewModel
  - There is a new view model for the notifications screen. It takes care of loading the follow requests and when a user deletes/accepts a notification takes care of the underlying logic.
  - There were modifications in the account repository for the addFriend and removeFriend logic. Now, when a user accepts a request , they can add themselves into the friend list of the other user in order to establish the two way connection. This also happens for the removal of the friend.
- Documentation
  - Kdoc comments added for the functions of the notitifcations and the modifications
- Tests
  - Added integration tests for the notification screen and notification view model and modified the other tests in order to take account of the logic changes.
- Manual verification:
  - I used two google accounts in order to create 2 accounts and try to friend each other.
  - It should work such that if you accept the friend request from one you should see yourself as a friend on the other.
  - If you delete the notification, you should not be a friend of the other person as well. 

# Why
This controls the users interactions and fulfills having their posts only be seen by their friends. 

# Screenshots / Videos

Screen with a notification:

image

Screen without a notification:

image

After a request is sent (I know the design is not well centered, I will fix this in a new PR)
image

After friending someone:
image

### DIFF
--- a/app/src/test/java/com/android/ootd/ui/notifications/NotificationViewModelTest.kt
+++ b/app/src/test/java/com/android/ootd/ui/notifications/NotificationViewModelTest.kt
@@ -165,7 +165,7 @@ class NotificationsViewModelTest {
     coEvery { mockNotificationRepository.getNotificationsForReceiver("user1") } returns
         listOf(notification)
     coEvery { mockUserRepository.getUser("sender1") } returns user
-    coEvery { mockAccountRepository.addFriend("user1", "sender1") } returns Unit
+    coEvery { mockAccountRepository.addFriend("user1", "sender1") } returns true
     coEvery { mockNotificationRepository.deleteNotification(notification) } throws
         Exception("Delete failed")
 


app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
```diff
@@ -191,12 +190,11 @@ class AccountRepositoryFirestore(private val db: FirebaseFirestore) : AccountRep
 
   override suspend fun addFriend(userID: String, friendID: String) {
     try {
-      // Instead of reading the friend's account (which we can't do due to privacy),
-      // check if the friend exists in the public User collection
       val friendUserDoc = db.collection(USER_COLLECTION_PATH).document(friendID).get().await()
 
       if (!friendUserDoc.exists()) {
-        throw NoSuchElementException("Friend with ID $friendID not found")
+        Log.e("AccountRepositoryFirestore", "The user with id ${friendID} not found")
```
bbianca2004: Logs are great for debugging, but please don't forget to remove them before merging.

app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
```diff
@@ -205,7 +203,21 @@ class AccountRepositoryFirestore(private val db: FirebaseFirestore) : AccountRep
...
+      try {
+        friendRef.update("friendUids", FieldValue.arrayUnion(userID)).await()
+      } catch (e: Exception) {
+        // If we can't update their account (maybe we're not in their friend list),
+        // log it but don't fail the entire operation
+        Log.w(
+            "AccountRepositoryFirestore",
+            "Could not add $userID to $friendID's friend list: ${e.message}")
+      }
     } catch (e: Exception) {
```
bbianca2004: Same thing here, don't forget to remove the logs, especially since they expose others user ids.

app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
```diff
@@ -205,7 +203,21 @@ class AccountRepositoryFirestore(private val db: FirebaseFirestore) : AccountRep
       // https://firebase.google.com/docs/firestore/manage-data/add-data , Update elements in an
       // array section
 
+      // Add the friendID in your friend list
       userRef.update("friendUids", FieldValue.arrayUnion(friendID)).await()
+
+      // Add yourself to the friendID's friend list.
+      // This works because we have not deleted the follow notification yet.
+      val friendRef = db.collection(ACCOUNT_COLLECTION_PATH).document(friendID)
+      try {
```
bbianca2004: I might be missing something, but it looks like this second update (adding the current user to the other user’s friendUids) could fail under the new Firestore rules(I think it is failing for me at least when I test your feature)
Since isFriend() now checks the other user’s friend list, wouldn’t this write be denied until that list already includes the current user?
Just wondering if this might be why the mutual friendship doesn’t persist after accepting a follow on my side, or if I’ve misunderstood the rule flow.

stefantaga24: No, so actually the update does go through in any case because there is a special rule in the firestore rules file which permits you to add yourself to their friend list only if there is a follow notification sent from them. This is tagged in the file with the following comment: // Non-owner can update ONLY to add themselves to friendUids if there's a follow notification. Unfortunately, the reason why it fails for you is because probably the accounts you tested on did not have the location added to them. Therefore your testing environment conflicts with the changes that Julien merged.

stefantaga24: I suggest you delete your two accounts from the account and user collection and re-login and try the feature if you can please :D

app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
bbianca2004: Ok, that makes sense!

firebase/firestore/firestore.rules
```diff
@@ -7,15 +7,24 @@ service cloud.firestore {
       return request.auth != null;
     }
 
-    // Mutual friendship check using friendUids lists on both users
+    // Checks whether the viewer is in the friend list of the author.
```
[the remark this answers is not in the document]
bbianca2004: You are totally right, thank you for the explanation, now it is clear that the issues when testing on my side are due to my accounts being outdated and not having the location field which leads to an error transforming document.

PR #208, opened 2025-11-03
feat(model): add notification repository

### LINKED ISSUE #225: Add notification repository
Add the underlying class that handles notifications between users. This class should be expandable such that we can add more than just follow notifications to it.

stefantaga24:
# Summary
Add the notifications repository for handling the notifications between users.

# What

- Model
  - The notification repository right now only implements follow notifications.
  - The follow notifications have a special uid which is of the form "{senderUid}_follow_{receiverUid}".
  - Any notification has the parameters: senderId, receiverId, type and content. Content is not used for the follow request but it might be important for new notifications that might be added.
- Documentation
  - Added Kdoc documents for the methods of the notifications repository.
- Tests
  - Added unit tests for the notifications repository in order to not clutter the instrumented tests.
- Firestore rules
  - The receiver and the sender can modify notifications as follow requests are implemented two way: if you add someone to your feed, they will automatically have you on their feed as well.
 
# Why
This makes the friend mechanism be more close to a social media app and the notification repository is a big building block for implementing the functionality.

Closes #225

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,63 +13,27 @@ on:
       - reopened
 
 jobs:
-  ci:
-    name: CI
-    # Execute the CI on the course's runners
-    runs-on: ubuntu-latest
-
+  # Job 1: Format Check (runs independently)
+  format-check:
+    name: Format Check
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    
     steps:
-      # First step : Checkout the repository on the runner
       - name: Checkout
         uses: actions/checkout@v4
         with:
           submodules: recursive
-          fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of Sonar analysis (if we use Sonar Later)
+          fetch-depth: 0
 
-      - name: CI Banner
-        run: |
-          echo '   ___   ___ _____ ___     ___ ___ '
-          echo '  / _ \ / _ \_   _|   \   / __|_ _|'
-          echo ' | (_) | (_) || | | |) | | (__ | | '
-          echo '  \___/ \___/ |_| |___/   \___|___|'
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/android/ootd/model/notifications/NotificationRepositoryFirestore.kt
```diff
@@ -0,0 +1,169 @@
...
+    } catch (e: Exception) {
+      Log.e(
+          "NotificationRepositoryFirestore",
+          "Error getting notifications for sender $senderId: ${e.message}",
+          e)
+      throw e
+    }
+  }
+
+  override suspend fun deleteNotification(notification: Notification) {
```
bbianca2004: NitPick: I see that the deleteNotification method intentionally doesn’t perform the actual delete,  makes sense if it’s being deferred to a future PR or maybe the actual deletion is not too important at the moment.
Though if you do not plan to change it any soon yourself, maybe consider renaming it temporarily something like `validateNotificationExistsForReceiver`, so whoever might work with it in the future doesn’t assume it already deletes from Firestore. Just an idea, it's up to you, let me know if I misunderstood anything.

bbianca2004 (approved):
Really great work on this PR ! The overall structure is very clean and easy to follow. I like how you separated the domain model, DTO, and repository interface, it keeps the architecture flexible and testable. The test coverage is also really good. It’s great to see such thorough unit tests, that really helps make the repository more robust. The Firestore rules are also well-aligned with the app’s logic and security model.

I just left two small comments (constants for field names and clarifying the intent behind deleteNotification) that are just small style or clarity suggestions. None of them are blockers , the functionality and overall design are solid. Great job!

PR #218, opened 2025-11-05
Fix : User can only write in his profile

Clemensito:
# Summary
Changed the rules to check that the uid matches the current users ID for write action. 

# What
The user has now a new parameter "ownerID" which eases the testing process, to verify that the creator of the user is the one modifying it, by default it is the uid. The implementation follows the same process as the account collection.
Also the check in the firebase rules that the ownerId is the same as the requesters ID 

- Tests
  - Updated the tests to follow the implementation of the new parameter "ownerId"

# Why
Necessary for integrity

### DIFF
--- a/app/src/androidTest/java/com/android/ootd/ui/searchscreen/UserSearchScreenTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/ui/searchscreen/UserSearchScreenTest.kt
@@ -81,7 +81,8 @@ class UserSearchScreenTest : FirestoreTest() {
     accountRepository.addAccount(
         Account(uid = currentUser.uid, username = secondUsername, ownerId = currentUser.uid))
 
-    userRepository.addUser(User(uid = currentUser.uid, ownerId = currentUser.uid,username = lastUsername))
+    userRepository.addUser(
+        User(uid = currentUser.uid, ownerId = currentUser.uid, username = lastUsername))
 
     // Input text to trigger dropdown
     composeTestRule


firebase/firestore/tests/setup.js
```diff
@@ -35,6 +36,16 @@ export async function setupFirestore(env, seed = {}) {
       await db.collection('accounts').doc(account.ownerId).set(account);
     }
 
+    // Users: derive friendUids from friendList and ensure ownerId exists
+    const users = seed.users ?? defaultUsers();
```
stefantaga24: I think here you need to code the function that actually creates the default users. Otherwise, you will get the error that the function is not defined. However, I am not sure if this is needed? You can just place an empty list considering that seed.users should always return something.

stefantaga24 (changes requested):
All in all it is a good PR! Small changes all around, just adding the ownerId such that a user cannot modify the data of other users. However, please make sure to fix the setup.js script such that the firestore rules tests can be run.

firebase/firestore/tests/firestore.rules.test.js
```diff
@@ -47,6 +47,20 @@ describe('Firestore rules - OOTD friend-only feed', () => {
     await assertFails(getDoc(doc(me, 'accounts/u3')));
   });
 
+    // users rules
+  test('Authenticated users can read any user doc but not write', async () => {
```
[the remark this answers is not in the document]
Clemensito: Thanks :)

firebase/firestore/tests/setup.js
Clemensito: Yup you're right. I replaced it with an empty string and now it works ! thanks for the hint

stefantaga24 (approved):
I see you changed the tests accordingly in order to add the ownerId. LGTM.

PR #222, opened 2025-11-05
Add the UI for the new camera functionalities

cocoStein:
# Summary
Add a safe image-orientation helper and polish the Camera UI flow + tests.

# What
- UI
  - `app/src/main/java/com/android/ootd/ui/camera/CameraScreen.kt`
    - Minor API cleanup: suppressed an intentional unused parameter (`onImageCaptured`) to remove noisy warnings.
    - Ensured capture flow uses the ViewModel to set captured image and show preview.
- ViewModel
  - No behavior changes to `CameraViewModel.kt` — it already sets UI state and triggers the preview correctly.
- Model / Utilities
  - Added `app/src/main/java/com/android/ootd/model/camera/ImageOrientationHelper.kt`
    - Loads a bitmap from a `Uri`, reads EXIF orientation, and returns a correctly rotated `Bitmap` as `Result<Bitmap>`.
    - Handles rotation (90/180/270°), flips, transpose/transverse, and recycles original bitmaps when appropriate. Protects against OOM.
- Tests
  - Added connected / instrumentation tests for camera preview flows and image orientation handling:
    - `app/src/androidTest/java/com/android/ootd/screen/CameraScreenPreviewTest.kt` (camera preview / preview-approve flow)
    - `app/src/androidTest/java/com/android/ootd/model/camera/ImageOrientationHelperConnectedTest.kt` (loads test image URIs, verifies orientation handling)
  - Existing Camera unit tests remain unchanged and pass.

# Why
- Real-world photos often contain EXIF orientation metadata; showing unrotated images leads to incorrect previews and poor UX. Adding a small, focused helper ensures captured images display correctly in the preview and when the user approves the photo.
- The suppression and small UI cleanup remove IDE noise and make tests more stable while preserving the intended flow where the ViewModel controls when the preview is shown.
- Adding connected tests improves confidence in camera + preview flows on device/emulator environments.

 > [!IMPORTANT]  
> COVERAGE NOTE - Zoom Slider (Lines 210-237 in CameraScreen.kt):
> The zoom slider UI is hardware-dependent and cannot be reliably covered in connected tests on emulators. The slider only renders when `maxZoomRatio > minZoomRatio`, which requires actual camera hardware with zoom capability. On some emulators runned locally, the tests works because the camera has a zoom features. But the emulator in the CI doesn't have a camera that has the avility to zoom (the test are failling stating that they don't see the slider). The zoom slider test function don't cover all of the lines so the coverage can't be fully set. I will ask the coaches what we can do with that.
> One fix could be to make the slider always visible, but i think it would look very bad, as it would not work and only display: `1x` every time

# Screenshots / Videos

This is an example of the actual flow of the new camera

[Screen_recording_20251106_144250.webm](https://github.com/user-attachments/assets/f2566bea-89bd-4037-a398-53fbf866843d)

This is an example of the slider only showing in camera that have the ability to zoom

[Screen_recording_20251106_184806.webm](https://github.com/user-attachments/assets/05d78932-a5f8-470e-a4a0-9636fdb7336b)

# Checklist
- [x] Tests: added/updated (connected tests for orientation and camera preview)
- [x] Documentation: KDoc added to `ImageOrientationHelper.kt`
- [x] Manual verification: app builds and camera flow tested locally (capturing → preview → approve)
- [x] Formatting: ran ktfmt for project formatting

### DIFF
--- a/app/src/main/java/com/android/ootd/model/camera/ImageOrientationHelper.kt
+++ b/app/src/main/java/com/android/ootd/model/camera/ImageOrientationHelper.kt
@@ -63,8 +63,16 @@ class ImageOrientationHelper {
       val exif = ExifInterface(inputStream)
       inputStream.close()
 
-      exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL)
-    } catch (e: IOException) {
+      val orientation =
+          exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL)
+
+      // Map ORIENTATION_UNDEFINED to ORIENTATION_NORMAL for proper handling
+      if (orientation == ExifInterface.ORIENTATION_UNDEFINED) {
+        ExifInterface.ORIENTATION_NORMAL
+      } else {
+        orientation
+      }
+    } catch (_: IOException) {
       ExifInterface.ORIENTATION_NORMAL
     }
   }
@@ -93,7 +101,9 @@ class ImageOrientationHelper {
         matrix.postRotate(-90f)
         matrix.postScale(-1f, 1f)
       }
-      else -> return bitmap // No rotation needed
+      ExifInterface.ORIENTATION_UNDEFINED,
+      ExifInterface.ORIENTATION_NORMAL -> return bitmap // No rotation needed
+      else -> return bitmap // No rotation needed for unknown orientations
     }
 
     return try {
@@ -106,7 +116,7 @@ class ImageOrientationHelper {
       }
 
       rotatedBitmap
-    } catch (e: OutOfMemoryError) {
+    } catch (_: OutOfMemoryError) {
       // If we run out of memory, return the original bitmap
       bitmap
     }
--- a/app/src/test/java/com/android/ootd/model/camera/ImageOrientationHelperTest.kt
+++ b/app/src/test/java/com/android/ootd/model/camera/ImageOrientationHelperTest.kt
@@ -45,12 +45,6 @@ class ImageOrientationHelperTest {
   // ========== Initialization Tests ==========
 
   @Test
-  fun `ImageOrientationHelper can be instantiated`() {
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/android/ootd/model/camera/ImageOrientationHelper.kt
```diff
@@ -0,0 +1,114 @@
...
+    return try {
+      val inputStream =
+          context.contentResolver.openInputStream(imageUri)
+              ?: return ExifInterface.ORIENTATION_NORMAL
+
+      val exif = ExifInterface(inputStream)
+      inputStream.close()
+
+      exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL)
+    } catch (e: IOException) {
```
asunluer: This e is redundant maybe you can put `_` so that you can get rid of the warning.

app/src/test/java/com/android/ootd/model/camera/ImageOrientationHelperTest.kt
```diff
@@ -0,0 +1,194 @@
...
+    unmockkAll()
+  }
+
+  // ========== Initialization Tests ==========
+
+  @Test
+  fun `ImageOrientationHelper can be instantiated`() {
+    val newHelper = ImageOrientationHelper()
+    assertNotNull(newHelper)
+  }
```
asunluer: This test seems redundant since the test below includes this test also you can remove it safely.

asunluer (changes requested):
I verified your implementation and I liked the way you coded. Your implementation was clear, easy to read and the documentation is very explicatif also. 

I added some suggestions there wasn't too much big issue but it is better to handle them in any case. 

Thank you so much for your contribution and your time. :) 

Additionally, please do not forget to fix the issues that Sonar Clous did mention ;)

asunluer (approved):
I checked it. Thank you so much for considering my suggestions. 

All is good ! Good job for implementing this ! :)

PR #245, opened 2025-11-07
fix(ui): implement only one color scheme for all themes

stefantaga24:
# Summary
Removed dark theme and hard coded color for navigation tab menu.

# What
Removed the dark theme and also hard coded the color for navigation menu. Therefore, I changed the color and theme files. 

- UI
  - Removed the colors from dark theme.
  - Added the new color for the bottom navigation bar selection.

# Why
We do not have a design for that and it leads to inconsistent screen behaviours on phones with it on. 

# Screenshots / Videos
image

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/theme/Theme.kt
+++ b/app/src/main/java/com/android/ootd/ui/theme/Theme.kt
@@ -6,7 +6,6 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.lightColorScheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.SideEffect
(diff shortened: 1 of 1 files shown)


MarcK0909 (changes requested):
LGTM but I find that when you are on a screen from the nav bar, its icon becomes barely visible due to it being a color close to white. I would suggest maybe using a darker color so that it's more visible. Also look at if you can remove the Sonar Cloud issue :)

stefantaga24:
All done! Changed to a darker lavender and removed the sonar issue :D

MarcK0909 (approved):
LGTM nice job :)

PR #250, opened 2025-11-07
Fix: Remove a bug where items were not displayed on release APK

asunluer:
# Summary

Fixes an issue where items were not displayed on Preview Screen after clicking the **Add Item** button in release APK. 

# What

- Data Layer (Model): 

   - Annotated data classes **Item**, **Material** and **ImageData** with `@Keep` to prevent ProGuard from removing them in release builds. 

# Why

When we build our app, Proguard might decide rename or remove certain fields on these data classes because they have to match with the serialized format. As a result, it causes the item to render after button click.  To prevent from this situation, adding`@Keep` annotation will preserve data classes as they are. 

# Screenshots / Videos

N/A - functional fix, no any visual changes

# Checklist
- [x] Tests: Check all existing tests pass
- [x] Manual verification: Run the app on release mode and check items are loading as intended to.


bbianca2004 (approved):
Great PR LGTM, thank you for fixing this bug!

PR #264, opened 2025-11-08
Feat: Add basic screen item inventory

cocoStein:
# Summary
Added a first iteration to the user inventory. At the moment the user can see a list of all the items he has previously posted and can edit them if needed.

> [!WARNING]
> As this changes the data class of the Account. In order to tests this you need to start with a fresh account so you have the new `itemsUids` field. And so that your inventory work.

# What
- Added a new field in the data class of the `Account`: `val itemsUids: List<String>`. This is used to store a list to all the items uids the user has ever posted.
- Added the corresponding function to the `AccountRepository`: `getItemsList(userID: String): List<String>`, `addItem(itemUid: String): Boolean`, `removeItem(itemUid: String): Boolean`

- Added a new `getItemsByIds(uuids: List<String>): List<Item>` in the `ItemRepository` to fetch multiple items at the same time (In batch of 10) to save computing time and ressources as we don't have to fetch them individually if we have too many.
- UI
  - New `InventoryScreen` that displays the items and can interact with them, can only edit them at the moment. 
    - `InventoryGrid` : new compose function that handle displaying the items in a grid pattern
    - `InventoryItemCard`: new compose to display a nice rounded square with the image of the item
- ViewModel
  - New View model for the inventory, handle the business logic of fetching the item from the firebase. At the moment the items are only cached, so they need to be refetched every time there's a changed. This will be implemented on a later PR.

As the inventory is using a lazy grid style. The items are fetch from the fb only if they are shown on the screen, so if you have many many items, you don't overload the db with requests as long as you don't scrol to far down. You can see an example in the second video.
- Tests
  - Unit tests for the new repository changes and the new inventory
  -  Integration tests for the inventory screen

# Why
This changes are made so the user can see in one space all the items he has, and later on he will be able to post a previously created item on a post. So he doesn't have to re create the items for every post

# Screenshots / Videos
This is an example of a user loading the inventory for the first time after killing the app, but having prior items posted on 2 different posts. This is why at first all of the items pictures are loading.

[Screen_recording_20251108_170752.webm](https://github.com/user-attachments/assets/c260d5d9-da60-4b3c-a9a7-4e5cb7808b9d)

This is an example of a video were you can see that the items are fetch from the db only if they appear on the screen.

[Screen_recording_20251109_205529.webm](https://github.com/user-attachments/assets/df1141ab-91fc-4516-987b-7ee743d893e8)

>[!NOTE]
> To test this new screen, you can go to the inventory tab in the center of your bottom navigation bar, if you have not posted any posts with some item yet, you'll see a text stating that you don't have any items.
> If you have posted some items, you'll see something like the video above

# Checklist

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/Inventory/InventoryScreen.kt
+++ b/app/src/main/java/com/android/ootd/ui/Inventory/InventoryScreen.kt
@@ -1,19 +1,8 @@
 package com.android.ootd.ui.Inventory
 
-import androidx.compose.foundation.background
-import androidx.compose.foundation.clickable
-import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.lazy.grid.GridCells
-import androidx.compose.foundation.lazy.grid.LazyVerticalGrid
-import androidx.compose.foundation.lazy.grid.items
-import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.CenterAlignedTopAppBar
 import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.ExperimentalMaterial3Api
@@ -29,16 +18,12 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.viewmodel.compose.viewModel
-import coil.compose.AsyncImage
-import com.android.ootd.model.items.Item
 import com.android.ootd.ui.navigation.NavigationActions
 import com.android.ootd.ui.navigation.Screen
 import com.android.ootd.ui.theme.Primary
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
```diff
@@ -43,6 +43,25 @@ class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsReposit
     return mapToItem(doc) ?: throw Exception("ItemsRepositoryFirestore: Item not found")
   }
 
+  override suspend fun getItemsByIds(uuids: List<String>): List<Item> {
+    if (uuids.isEmpty()) return emptyList()
+
+    val ownerId = Firebase.auth.currentUser?.uid ?: throw Exception(NOT_LOGGED_IN_EXCEPTION)
+
+    // Batch of 10 item for each querries
```
stefantaga24: Would this be useful for like lazy loading? Instead of loading everything into memory you can like load only when the user scrolls? Not to be implemented right now but just curious.

app/src/main/java/com/android/ootd/ui/Inventory/InventoryScreen.kt
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * Individual item card displaying a small rounded square with the item's image.
+ *
+ * @param item The item to display
+ * @param onClick Callback when the item is clicked
+ * @param modifier Modifier for styling
+ */
+@Composable
+private fun InventoryItemCard(item: Item, onClick: () -> Unit, modifier: Modifier = Modifier) {
```
stefantaga24: I would honestly place this in another file like InventoryItemCard.kt , it would be easier to read :D

app/src/main/java/com/android/ootd/ui/post/AddItemsViewModel.kt
```diff
@@ -177,8 +180,17 @@ open class AddItemsViewModel(
                 link = state.link,
                 ownerId = ownerId)
 
-        _addOnSuccess.value = true
         repository.addItem(item)
+
+        // Add item to user's inventory
+        val addedToInventory = accountRepository.addItem(itemUuid)
```
stefantaga24: Here, if the item is not added to the inventory I think you should fail the entire operation because otherwise the item will never be added to the inventory and you are stuck in an illegal state where you have an item posted but not in the inventory. If the entire operation fails it is not that bad for the user, they can just press one button and retry. Tell me what you think :D

app/src/main/java/com/android/ootd/ui/post/EditItemsViewModel.kt
```diff
@@ -196,6 +200,13 @@ open class EditItemsViewModel(
     viewModelScope.launch {
       try {
         repository.deleteItem(state.itemId)
+
+        // Remove item from user's inventory
+        val removedFromInventory = accountRepository.removeItem(state.itemId)
```
stefantaga24: Here again the same thing as above, I am wondering if maybe we should fail the operation?

app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
```diff
@@ -43,6 +43,25 @@ class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsReposit
     return mapToItem(doc) ?: throw Exception("ItemsRepositoryFirestore: Item not found")
   }
 
+  override suspend fun getItemsByIds(uuids: List<String>): List<Item> {
+    if (uuids.isEmpty()) return emptyList()
+
+    val ownerId = Firebase.auth.currentUser?.uid ?: throw Exception(NOT_LOGGED_IN_EXCEPTION)
```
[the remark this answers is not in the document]
cocoStein: thanks ! I want to make the exception clearer so we have an idea of why something doesn't work

app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
cocoStein: I confirm that it already works as expected, i artifcially increased the padding to make them appear outside the screen. and at first on the screen we only have 9 elements. And you can see that the last 2 are loaded when we start seeing them 🔥 🚀 

[Screen_recording_20251109_205529.webm](https://github.com/user-attachments/assets/59cd6c0f-238a-4f95-8d69-7eb171e47744)

app/src/main/java/com/android/ootd/ui/post/AddItemsViewModel.kt
```diff
@@ -177,8 +180,19 @@ open class AddItemsViewModel(
...
+        if (!addedToInventory) {
+          repository.deleteItem(itemUuid)
+          FirebaseImageUploader.deleteImage(uploadedImage.imageId)
+          setErrorMsg("Failed to add item to inventory. Please try again.")
```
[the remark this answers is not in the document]
stefantaga24: Yes I think this is the correct way to do it, we need to rollback what we added in the first place and let the user try again. I don't think this is necessarily duplicated code as you are just rolling back the operation.
You can resolve this conversation if you think this is okey :D

PR #271, opened 2025-11-09
fix(debug_apk): Remove signing with release keys

stefantaga24:
# Summary
Changed github workflow to remove signing with release key.

# What
Changed github workflow to remove signing with release key.

# Why
Signing with the wrong key prevents the user from signing in.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -40,14 +40,6 @@ jobs:
       - name: Set up Android SDK
         uses: android-actions/setup-android@v3
 
-      - name: Decode signing and optional files
-        shell: bash
-        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-release.yml
cocoStein: Hey if we don't use the KEY, i don't think we need this task anymore ?

tell me what you think

stefantaga24: I think it is worth to keep it just to have both the debug and release , for testing purposes so we can compare when something on release does not work

stefantaga24: I was thinking about the wrong thing, I will take it out

PR #297, opened 2025-11-11
Feat: Added feature to see others' items on their post

asunluer:
# Summary

This PR introduces `SeeFitScreen.kt` file implementation. It is a new UI feature which shows the items of the users that they wore on that day, by clicking on **See Fit** button from feed. 

# What

- UI
  - `SeeFitScreen` is implemented 
  - `SeeFitCard`, `SeeFitGrid` and `SeeFitItemDialog` components are implemented. 
- Documentation
  - Kdoc comments added for each composable functions
- Tests
  - Integration tests added :
      - verifies the existing of the components `SeeFitCard`, `SeeFitGrid` and `SeeFitItemDialog`
      - checks the existing of the fields to be displayed
      - checks if items of related post is uploading correctly. 

# Why

Sometimes when we see a cloth that we liked, we ask to our friends where did they get it. Since, this app introduces what we are wearing on that day, instinctively, we want to find out where the item comes from. Hence, this screen is implemented for the users who want to buy the same item or get an inspiration for their look.

# Screenshots / Videos

https://github.com/user-attachments/assets/edc2122b-9c06-4933-8909-aac3a4c5a6e1

# Checklist
- [x] Tests: added as appropriate
- [x] Documentation: KDocs added for each composable functions. 
- [x] Manual verification: The screen is displayed as shown in the video. 
- [x] Manual verification : You can go back from see fit to feed screen

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/feed/SeeFitScreen.kt
+++ b/app/src/main/java/com/android/ootd/ui/feed/SeeFitScreen.kt
@@ -3,12 +3,15 @@ package com.android.ootd.ui.feed
 import android.widget.Toast
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.outlined.ArrowBack
 import androidx.compose.material3.CenterAlignedTopAppBar
+import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/ootd/ui/feed/SeeFitScreen.kt
```diff
@@ -1,18 +1,139 @@
...
+    seeFitViewModel: SeeFitViewModel = viewModel(),
+    postUuid: String = "",
+    goBack: () -> Unit = {}
+) {
+
+  val context = LocalContext.current
+  val uiState by seeFitViewModel.uiState.collectAsState()
+  val items = uiState.items
+
+  LaunchedEffect(Unit) { seeFitViewModel.getItemsForPost(postUuid) }
```
bbianca2004: I just have a question, if we use Unit here it means this block can re-run on every recomposition even if the postUuid hasn’t changed? Shouldn't it be scoped out to the postUuid so the data will only be fetched again when navigating to a different post's items? As in `LaunchedEffect(postUuid)` instead of Unit.

app/src/main/java/com/android/ootd/ui/feed/SeeFitScreen.kt
```diff
@@ -1,18 +1,139 @@
...
+                    navigationIconContentColor = MaterialTheme.colorScheme.onSurfaceVariant),
+        )
+      },
+  ) { innerPadding ->
+    if (items.isEmpty()) {
+      Column(
+          modifier = Modifier.fillMaxSize().padding(innerPadding),
+          verticalArrangement = Arrangement.Center,
+          horizontalAlignment = Alignment.CenterHorizontally) {
+            Text(
```
bbianca2004: You’re showing the “No items” message any time the list is empty which is great but would that also happen even before data finishes loading especially when connection is slow? This causes a flash of the empty state before content appears.
Maybe you could consider adding a loading flag to the uiState (uiState.isLoading) and show a CircularProgressIndicator() until the fetch completes, what do you think??

bbianca2004 (changes requested):
I tested your feature and everything works really smoothly. The SeeFit flow feels complete now, the cards look great and the overall layout fits nicely with the rest of the app. You did a solid job keeping the UI modular (ItemCard, ItemGridScreen, SeeItemDetailsDialog), and the test coverage is impressive, well done for that!
I’ve left a few small remarks though with someimprovements , please let me know what you think about it :))

app/src/main/java/com/android/ootd/ui/feed/SeeFitScreen.kt
asunluer: You are totally right on that if I used the same screen again for different **potsUuid**s. However, when you look at the navigation I push and pop the screen every time I click on **See Fit** and **go back** button. This means that every time I navigate to SeeFitScreen it creates a new screen and the compose navigation pıshes the new destination onto the stack. And when I go back, it pops the screen which destroys the entire compositions. Thus, it is safe to use `LaunchedEffect(Unit)`. And if you try with the emulator also you can see for the different **items** for different **postUuid**s.

asunluer:
Thank you so much for your reviews. It is nice to discuss with you. They were valuable comments :)

PR #299, opened 2025-11-11
Feat: Inventory part 3 - add item from inventory on posts

cocoStein:
# Summary
This is the 3rd iteration of the inventory screen. In this version, I added the ability for a user to pick items from his own inventory when he wants to choose which item will be associated with his current post.

# What
- UI
  - Screen: `SelectInventoryItemScreen` it takes the same custom composable element as the inventory to display each items sorted by their categories
  - New `AlertDialog` when the user click `Add item` on the `PreviewPostScreen` so the user can choose from selecting from the inventory or creating a new item.
  -   - New assets/resources added (icons, drawables, colors)
- ViewModel
  - State changes or new state fields
  - Business logic / repository / API interactions modified
  - Navigation or side-effect changes
- Tests
  - Fixed the old implementation of the test / navigation to follow the new logic
  - Added new Connected test to test my new screens and VM
> [!NOTE]
> I chose to  use a `AlertDialog` to select between creating a new item or by choosing one from the inventory because that's what makes the most sense for me. We will have to do UX test to see if that's the best choice.

# Why

# Screenshots / Videos
Video showing an example of the new complete flow of adding items on a post

[Screen_recording_20251111_205444.webm](https://github.com/user-attachments/assets/010aecd9-3503-468e-9592-ad19ac70cb22)

This is an example of what the "preview" inventory looks like. It's basically the same screen as the inventory except that when you click on an items you select it rather that edit it.
Screenshot_20251111_205609

This is an example of the pop up that shows when you press the "add item" button
Screenshot_20251111_205603

# Checklist

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/navigation/NavigationActions.kt
+++ b/app/src/main/java/com/android/ootd/ui/navigation/NavigationActions.kt
@@ -84,9 +84,9 @@ sealed class Screen(
   }
 
   data class SelectInventoryItem(val postUuid: String) :
-      Screen(route = "selectInventoryItem?postUuid=$postUuid", name = "Select Inventory Item") {
+      Screen(route = "selectInventoryItem/$postUuid", name = "Select Inventory Item") {
     companion object {
-      const val route = "selectInventoryItem?postUuid={postUuid}"
+      const val route = "selectInventoryItem/{postUuid}"
     }
   }
 }
--- a/app/src/main/java/com/android/ootd/ui/post/SelectInventoryItemViewModel.kt
+++ b/app/src/main/java/com/android/ootd/ui/post/SelectInventoryItemViewModel.kt
@@ -55,6 +55,13 @@ class SelectInventoryItemViewModel(
    * @param postUuid The UUID of the post to add items to
    */
   fun initPostUuid(postUuid: String) {
+    // Guard for blank/empty postUuid
+    if (postUuid.isBlank()) {
+      _uiState.value =
+          _uiState.value.copy(isLoading = false, availableItems = emptyList(), errorMessage = null)
+      return
+    }
+
     if (_uiState.value.postUuid == postUuid) return
 
     _uiState.value = _uiState.value.copy(postUuid = postUuid)
--- a/app/src/androidTest/java/com/android/ootd/navigation/NavigationTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/navigation/NavigationTest.kt
@@ -5,10 +5,12 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
 import androidx.navigation.NavHostController
+import androidx.navigation.NavType
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.navigation
 import androidx.navigation.compose.rememberNavController
+import androidx.navigation.navArgument
 import com.android.ootd.ui.map.MapScreen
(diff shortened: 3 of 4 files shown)


app/src/main/java/com/android/ootd/ui/post/SelectInventoryItemViewModel.kt
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * Loads items from the user's inventory, excluding items already associated with the current
+   * post.
+   */
+  private fun loadAvailableItems() {
+    val postUuid = _uiState.value.postUuid
+    if (postUuid.isEmpty()) {
+      Log.w("SelectInventoryItemViewModel", "loadAvailableItems called with empty postUuid")
+      return
```
MarcK0909: Should we maybe also change the isLoading state before returning like this if it was previously true : 
if (postUuid.isBlank()) {
_uiState.value = _uiState.value.copy(isLoading = false, availableItems = emptyList())
return
}

app/src/main/java/com/android/ootd/ui/post/SelectInventoryItemViewModel.kt
```diff
@@ -0,0 +1,138 @@
...
+        val allItems = itemsRepository.getItemsByIds(itemIds)
+
+        // Get items already associated with this post
+        val postItems = itemsRepository.getAssociatedItems(postUuid)
+        val postItemIds = postItems.map { it.itemUuid }.toSet()
+
+        // Filter out items already in the post
+        val availableItems = allItems.filter { it.itemUuid !in postItemIds }
+
+        _uiState.value = _uiState.value.copy(availableItems = availableItems, isLoading = false)
```
MarcK0909: Should we maybe clear the messages bc the operation is succesful.

app/src/main/java/com/android/ootd/ui/navigation/NavigationActions.kt
```diff
@@ -82,6 +82,13 @@ sealed class Screen(
       const val route = "editItem/{itemUid}"
     }
   }
+
+  data class SelectInventoryItem(val postUuid: String) :
```
MarcK0909: This mixes an instance-specific route string (with postUuid interpolated) and a companion constant route for registration. Using a query format with interpolation may be confusing and error-prone; name duplication (route constant and property) can cause confusion. I would suggest : companion object { const val ROUTE = "selectInventoryItem/{postUuid}" }
usage: Screen.SelectInventoryItem(postUuid).route should build "selectInventoryItem/$postUuid"
register composable with route = SelectInventoryItem.ROUTE and argument navArgument("postUuid") { type = NavType.StringType }

MarcK0909 (changes requested):
Very good work and great idea to lazily load the items it's much better this way and looks very nice thanks for your work.

app/src/main/java/com/android/ootd/ui/post/SelectInventoryItemViewModel.kt
```diff
@@ -0,0 +1,138 @@
...
+
+  private val _uiState = MutableStateFlow(SelectInventoryItemUIState())
+  val uiState: StateFlow<SelectInventoryItemUIState> = _uiState.asStateFlow()
+
+  /**
+   * Initializes the ViewModel with the post UUID and loads available items.
+   *
+   * @param postUuid The UUID of the post to add items to
+   */
+  fun initPostUuid(postUuid: String) {
```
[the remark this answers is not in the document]
MarcK0909: I agree it's very unlikely to happen but an extra safeguard might avoid us some dumb mistakes plus it's very easy to do, but I also see your point so feel free to add it or not

PR #310, opened 2025-11-12
feat(test): add second end to end test

stefantaga24:
# Summary
Added second end 2 end test.

# What

- Tests
  - Added a new end 2 end test
  - We test the search function and the follow function in the second end 2 end test compared to the first
  - We also check the posting flow and the adding of an item flow without actually using the camera as that is not supported yet. 
  - We also test the inventory to see that the add item button appears and when there are no items it loads correctly
  - Navigation through the bottom bar is also tested when going to the inventory.
  - Added better abstractions for testing as it can cut line length by 75% especially when using waitUntil for checking elements being displayed.

# Why

This second end to end test is more comprehensive and integrates more of the functionalities of the app.

### DIFF
--- a/app/src/androidTest/java/com/android/ootd/screen/RegisterScreenTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/screen/RegisterScreenTest.kt
@@ -22,6 +22,7 @@ import com.android.ootd.ui.map.LocationSelectionTestTags
 import com.android.ootd.ui.register.RegisterScreen
 import com.android.ootd.ui.register.RegisterScreenTestTags
 import com.android.ootd.ui.register.RegisterViewModel
+import com.android.ootd.utils.verifyElementDoesNotAppearWithTimer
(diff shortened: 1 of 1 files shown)


cocoStein (changes requested):
First of all, I really like how you created a function for every piece of test and that you made a file for the mocking. This will make e2e 3 easier and it will also make the process of fixing the test when something changes way easier. And the test is easy to follow now. Good job. 

On another side, i think now that we have our own custom camera screen, you can use it to take actually pictures for both the items and the posts. If you can try if it works, it would make this test perfect. 

PS: don't forget to grant the right permission for the camera i don't think you'll be able to emulate that as it's an android native (i could not manage to test this part).

stefantaga24:
Yes, I will try to only bypass the camera as the other things should be able to work. Practically, I will not take into account that there are no images added because they cannot be produced yet.

PR #318, opened 2025-11-12
Fix: Fixed a bug where items of the user's friends were not displayed as intended

asunluer:
# Summary

This PR introduces a fixed bug issue on fetching items of the user's friend. Before, the user couldn't see the items of their friends even they posted. 

# What

- Repository 
   -  In `feedRepositoryFirestore`, `getPostById` is implemented which takes users postId and returns the post's data. 
   - In `feedRepositoryFirestore`, a private function `mapToPost` is implemented which maps documents to `OutfitPost` data class. 
   - In `itemRepositoryFirestore`,  `getFriendItemsForPost` is implemented which takes the **postUuid** and **ownerId** - friend's id and returns the **list of items**.  
- ViewModel
  - `feedRepository: FeedRepository` added as argument in the view model 
  -  New implemented functions fetches the user's items in `getItemsForPost` function. 
     - First, we get the friend's or user's post data. 
     - Then using the **ownerId** of that post, we retrieve the corresponding items by calling `getFriendItemsForPost(postUuid, friendId)`
   - Firebase Rules: Missing permission for items on friend's side is added. 
- Documentation
  - Kdoc comments updated
- Tests
  -  `SeeFitViewModelTest` unit test is updated. 
       - Refactored tests with respect to new functions (getItemsForPost and getFriendItemsForPost)
  - Integration tests updated : 
      - `FeedRepositoryFirestoreTest` test is updated, new functions are added to verify the correctness of `getPostById`
      - `ItemsRepositoryFirestoreTest` is updated based on the new implemented function `getFriendItemsForPost` to check its correctness. 
      - `FeedScreenTest` updated because it uses a new repository which needs to have new function in that fake repository
      - `PreviewItemScreenTest` updated, it is the same reasoning with the `FeedScreenTest`
      - `SelectIncventoryItemScreenTest` updated because of the same reasons.
      - `InMemoryItem.kt` updated because of change of the `ItemRepositoryFirestore` 

# Why

Before when the user tries to see the posted items of their friend, they couldn't see it because of missing permission for the items. Because, before only the authenticated user can see their items. And I added a condition that tells only the friends of that user can see the items. However, this didn't solve the bug quickly because in `getAssociatedItems` we were fetching only the authenticated user's items because we were checking only the ownerId of the authenticated user. Since it was not associated with its friend's **ownerId** it discarded them even we passed a **postUuid** argument. To fix this issue, I created two functions, one it was getting the whole post's data from the database because we need to use the specified post's **ownerId** to fetch the corresponding post with its items. Then by using this ownerId with the postId because each owner can have multiple posts, it extracts the list of items of that user. This is how the bug is fixed. 

# Screenshots / Videos

# Checklist
- [x] Tests: added and updated  as appropriate
- [x] Documentation: Documentations are added.
- [x] Manual verification: the items are fetched correctlt from **SeeFit** Screen

### DIFF
--- a/app/src/androidTest/java/com/android/ootd/model/items/ItemsRepositoryLocalTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/model/items/ItemsRepositoryLocalTest.kt
@@ -425,6 +425,136 @@ class ItemsRepositoryLocalTest {
     TestCase.assertTrue(repository.hasItem(item1.itemUuid))
   }
 
+  @Test
+  fun getFriendItemsForPostReturnsEmptyWhenNoMatch() = runTest {
+    val postUuid = "post-with-no-friend-items"
+    val friendId = "friend-user-789"
+
+    // Add items for current user only
+    repository.addItem(item1.copy(postUuids = listOf(postUuid), ownerId = "user-123"))
+
+    val friendItems = repository.getFriendItemsForPost(postUuid, friendId)
+
+    TestCase.assertTrue(friendItems.isEmpty())
+  }
+
+  @Test
+  fun getFriendItemsForPostFiltersCorrectlyByPostUuid() = runTest {
+    val postUuid1 = "friend-post-aaa"
+    val postUuid2 = "friend-post-bbb"
+    val friendId = "friend-999"
+
+    // Friend items for post1
+    val friendItem1 =
+        item1.copy(itemUuid = "friend-item-1", postUuids = listOf(postUuid1), ownerId = friendId)
+    val friendItem2 =
+        item2.copy(itemUuid = "friend-item-2", postUuids = listOf(postUuid1), ownerId = friendId)
+
+    // Friend items for post2
+    val friendItem3 =
+        item3.copy(itemUuid = "friend-item-3", postUuids = listOf(postUuid2), ownerId = friendId)
+
+    repository.addItem(friendItem1)
+    repository.addItem(friendItem2)
+    repository.addItem(friendItem3)
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/ootd/model/feed/FeedRepositoryFirestore.kt
```diff
@@ -124,4 +125,35 @@ class FeedRepositoryFirestore(private val db: FirebaseFirestore) : FeedRepositor
   override fun getNewPostId(): String {
     return java.util.UUID.randomUUID().toString()
   }
+
+  override suspend fun getPostById(postUuid: String): OutfitPost? {
```
bbianca2004: Thank you for this addition, this was quite necessary considering that beforehand there was no function to fetch a post by its id, which I feel is quite fundamental.

bbianca2004 (approved):
Thank you so much for fixing this issue, it is a really significant part of our app and the bug was really not an easy one to spot, especially considering it was involved with the rules and there was no permission denied warning hinting at what the underlying issue is. I also tested it, all works well!!
I see that you have also added quite comprehensive tests too, good job on that, I feel like some of them could be removed since they are testing the same logic, but for now it is much better to leave them in. 
Overall great work on this PR, you really took on a big task for fixing this and you delivered big time :))

asunluer:
Thank you so much for your review. I appreciate it a lot. And thank you so much to stay awake and help me to finish my PR. You are a really good, helpful teammate. I am lucky to have you :)

PR #356, opened 2025-11-17
Add push notifications for the follow notifications

stefantaga24:
# Summary
Have push notifications appear while the user is on the app. The notifications appear only if the user has enabled them.

# What
Added a listener for notifications and a simple notification for the follow request.

- UI
  - Added a specific button that lets the user enable the notifications on their own accord.
  - Created a small notification pop up for the user. It has no interactions yet.
- ViewModel
  - Added a new notification listener in the MainActivity and it starts listening when the app launches.
- Documentation
  - Updated Kdoc comments for the new code added in MainActivity and the NotificationRepositoryFirestore

# Why

It is worth it to make the app more dynamic, such that when the user enters, he shouldn't have to always check for new notification requests. In the future, if we implemented a listener that works even when the app is closed, we will be able to make a truly immersive experience. 

# Screenshots / Videos

When somebody wants to follow you , you should get this following notification. I have used the fact that we can follow ourselves (which helps a lot with debugging, I will remove it in the next version of the app). 

image

# Checklist
- [X] Tests: Need to add tests for the notification listener. 
- [X] Documentation: Need to add documentation for the new methods.

### DIFF
--- a/app/src/main/java/com/android/ootd/MainActivity.kt
+++ b/app/src/main/java/com/android/ootd/MainActivity.kt
@@ -196,6 +196,12 @@ fun OOTDApp(
   fun observeUnpushedNotifications(userId: String) {
     if (listenerRegistration != null) return
 
+    if (testMode) {
+      sendLocalNotification(
+          Notification(
+              uid = "", senderId = "", receiverId = "", type = "", content = "", wasPushed = false))
+    }
+
     listenerRegistration =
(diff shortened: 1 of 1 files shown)


stefantaga24:
I think I will add it in a new PR to be honest :D

cocoStein (approved):
Just fix the coverage and everything will be fine 🔥

PR #362, opened 2025-11-19
Feat: Implement likes logic

### LINKED ISSUE #359: Implement likes logic

bbianca2004:
# Summary
This PR implements the logic for post likes, including repository implementation, security rules, and integrated firebase emulated tests for the CRUD operations.

# What

- ViewModel / Data / Repository
  - Introduced a `Like` data model (`postId`, `likerUserId`, `timestamp`) 
  - Implemented `LikesFirestoreRepository` that stores likes under  
    `likes/{postId}/users/{likerUserId}` with basic CRUD operations:
    - `likePost(...)`
    - `unlikePost(postId, userId)`
    - `isPostLikedByUser(postId, userId)`
    - `getLikesForPost(postId)`
    - `getLikeCount(postId)`
  - Added `LikesRepository` interface and `LikesRepositoryProvider` to expose the Firestore implementation to the rest of the app.

- Firestore rules
  - Added dedicated security rules post likes 
  - For now, only your friends can like/unlike your posts, but friends of a user that are not mutual friends can see each others likes on the user's post

- Tests
  - Added `LikesFirestoreRepositoryTest` (instrumented, using the Firebase emulator) to cover the repository paths:
    - `likePost_createsDocument`
    - `unlikePost_deletesDocument`
    - `isPostLikedByUser_returnsCorrectState`
    - `getLikeCount_returnsCorrectValue`
  - Tests create a post document and verify likes are written/removed under `likes/{postId}/users/{userId}` as expected.

# Why

The likes feature is the first in the series of reactions that users can leave on each others posts. This feature is crucial to the social aspect of our app and to increase user engagement. This PR is the first step of this new feature, as it implements and exposes the logic of the likes feature.

# Checklist
- [x] Tests: added/updated instrumented tests for `LikesFirestoreRepository`
- [x] Documentation: Added Kdoc for the Likes Repository
- [x] Manual verification: ran app against emulator, created a post, and confirmed likes documents are created/removed under `post_likes/{postId}/users/{userId}`

Closes #359

### DIFF
--- a/app/src/androidTest/java/com/android/ootd/model/post/LikesFirestoreRepositoryTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/model/post/LikesFirestoreRepositoryTest.kt
@@ -125,7 +125,6 @@ class LikesFirestoreRepositoryTest : FirestoreTest() {
 
     val likes = repo.getLikesForPost(postId)
 
-    assertEquals(1, likes.size)
     val like = likes.first()
     assertEquals(postId, like.postId)
     assertEquals(userA, like.postLikerId)


app/src/main/java/com/android/ootd/model/posts/LikesFirestoreRepository.kt
```diff
@@ -0,0 +1,92 @@
...
+          db.collection(LIKE_COLLECTION_PATH)
+              .document(postId)
+              .collection("users")
+              .document(userId)
+              .get()
+              .await()
+
+      snap.exists()
+    } catch (e: Exception) {
+      Log.w("LikeRepository", "Failed to check like state for $postId", e)
```
asunluer: I realised that you used `Log.w` but since this part catches an exception it should be better that you used `Log.e` . There is a StackOverflow page that explains their differences also. You can find [here](https://stackoverflow.com/questions/7959263/android-log-v-log-d-log-i-log-w-log-e-when-to-use-each-one).

app/src/main/java/com/android/ootd/model/posts/Like.kt
```diff
@@ -0,0 +1,3 @@
+package com.android.ootd.model.posts
+
+data class Like(val postId: String = "", val likerUserId: String = "", val timestamp: Long = 0L)
```
asunluer: I have suggestion for the naming what do you think about `likedPostUserId`, `userLikedPost`. What do you think of them ?

app/src/main/java/com/android/ootd/model/posts/LikesFirestoreRepository.kt
```diff
@@ -0,0 +1,92 @@
...
+      Log.e("LikeRepository", "Failed to fetch likes for $postId", e)
+      emptyList()
+    }
+  }
+
+  override suspend fun getLikeCount(postId: String): Int {
+    return try {
+      db.collection(LIKE_COLLECTION_PATH).document(postId).collection("users").get().await().size()
+    } catch (e: Exception) {
+      Log.w("LikeRepository", "Failed to retrieve like count for $postId", e)
```
asunluer: Same here as above :)

app/src/main/java/com/android/ootd/model/posts/LikesFirestoreRepository.kt
```diff
@@ -0,0 +1,92 @@
...
+  override suspend fun unlikePost(postId: String, userId: String) {
+    try {
+      db.collection(LIKE_COLLECTION_PATH)
+          .document(postId)
+          .collection("users")
+          .document(userId)
+          .delete()
+          .await()
+    } catch (e: Exception) {
+      Log.e("LikeRepository", "Failed to remove like for post $postId", e)
```
asunluer: Since, you are using `LikeRepository` as a tag in different functions. Instead of repeating over and over again, what do you think about doing them as 


```suggestion
      private const val TAG_LIKE_REPOSITORY = LikeRepository
```

app/src/androidTest/java/com/android/ootd/model/post/LikesFirestoreRepositoryTest.kt
```diff
@@ -0,0 +1,175 @@
...
+    repo.likePost(
+        Like(postId = postId, likerUserId = userA, timestamp = System.currentTimeMillis()))
+
+    val likes = repo.getLikesForPost(postId)
+
+    assertEquals(1, likes.size)
+    val like = likes.first()
+    assertEquals(postId, like.postId)
+    assertEquals(userA, like.likerUserId)
+  }
```
asunluer: Since these two tests check the similar situations it should be better to maybe merge them rather than having two separate tests. Thus, we have less integrated tests for CI to avoid run. slower. What do you think about it ? 

Also, it can be better to have maybe more than one `Like` data because name can mislead a little bit. I would expect to have more likes for example.

app/src/androidTest/java/com/android/ootd/model/post/LikesFirestoreRepositoryTest.kt
```diff
@@ -0,0 +1,175 @@
...
+
+  @Test
+  fun getLikesForPost_returnsEmptyListOnError() = runTest {
+    FirebaseEmulator.auth.signInAnonymously().await()
+
+    val likes = repo.getLikesForPost("bad/id")
+
+    assertTrue(likes.isEmpty())
+  }
+}
```
asunluer: Since the **postId** will be ignored by the Firestore because of its naming this will throw an error, you can use the last two line for `likePost_throwsonFirestoreFailure()` test. What do you think ?

asunluer (changes requested):
Overall your PR is so good, clear enough and very expressive. Maybe, you can mention also that there will be following implementations/changes based on this PR for your next tasks. 

Additionally, the name of your collection is too nested, it makes harder for me to get the idea but if you think this is the easiest and correct way to have security rules, you can leave it like that. 

You did a really good job ! Your comments are very explicatif they help me to understand the functionalities. 

Very good PR ! Thank you so much for your contribution.

app/src/main/java/com/android/ootd/model/posts/LikesFirestoreRepository.kt
bbianca2004: Yes, you are totally right, thank you for spotting that, it was a small error on my part :))

app/src/main/java/com/android/ootd/model/posts/LikesFirestoreRepository.kt
```diff
@@ -0,0 +1,92 @@
...
+      db.collection(LIKE_COLLECTION_PATH)
+          .document(like.postId)
+          .collection("users")
+          .document(like.likerUserId)
+          .set(
+              mapOf(
+                  "postId" to like.postId,
+                  "likerUserId" to like.likerUserId,
+                  "timestamp" to like.timestamp),
+              SetOptions.merge())
```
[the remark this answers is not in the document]
bbianca2004: Yes, I agree with you, I am overwriting all these fields so there is no need for merge.

firebase/firestore/firestore.rules
```diff
@@ -132,6 +132,42 @@ service cloud.firestore {
...
+                ||
+                // Or a friend of the owner
+                isFriend(
+                  request.auth.uid,
+                  get(/databases/$(database)/documents/posts/$(postId)).data.ownerId
+                )
+             );
+
+      allow delete: if isAuthed()
+          && request.auth.uid == userId;
```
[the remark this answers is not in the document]
bbianca2004: I feel like a user should always be able to delete their like even after the owner of the post unfriends them, every action should be able to have an undo.  A user who wasn't friends with the post owner in the first place could not have liked their post anyway, so they wouldn't have anything to delete.

app/src/androidTest/java/com/android/ootd/model/post/LikesFirestoreRepositoryTest.kt
bbianca2004: Yes, with that I totally agree, will do as you said, thanks!

asunluer (approved):

PR #373, opened 2025-11-19
Feat: Apply image compressor

asunluer:
# Summary

This PR introduces that image compressor is applied before we upload images to **Firebase Storage** and retrieves these images from there. 

# What

- UI
  - Updated screens: 
     - `AddItemScreen`
     - `EditItemScreen`
     - `PreviewtemScreen`
- ViewModel
  - Applied `compressImage` function to modified files: 
     - `EditItemViewModel`
     - `AddItemViewModel`
     - `OutfitPostViewModel
- Documentation
  - Kdoc comments added/updated
  - External docs (README updates, wiki) updated
- Tests
  - Unit tests updated : 
     - `AddItemsViewModelTest`
     - `EditItemsViewModelTest` 
  - Integration tests updated : 
      - `FirebaseImageUploaderTest`
      - `OutfitPostFirebaseRepositoryTest`
      - `AddItemsFirebaseRepositoryTest`
      - `EditItemsFirebaseRepositoryTest`
      - `OutfitPreviewViewModelFirebaseTest`
  - UI tests updated: 
     - `AddItemsScreenTest`
     -  `EditItemsScreenTest`

# Why

This PR is introducing the image compressor application on several parts of the code before uploading the images to **Firebase Storage**.  Since, the user will upload lots of images to post and see others post every day, we need to keep the storage as empty as possible. So, we decided to apply this functionality in the code. 

# Screenshots / Videos

Before : The size of file is 38 000 000 bytes

After: Retrieved from firebase storage and the file size is 833,928 bytes
isMMWcB7Ff7AiuDwE009

# Checklist
- [x] Tests: added unit, integrated tests.
- [x] Tests: updated unit, integrated tests. 
- [x] Documentation: inline comments are added
- [x] Manual verification: check if the existing functionalities work. 
- [x] Manual verification: check if the image compressor compress images correctly

# Future work 
- There exists some functions that are not used anymore but I want to discuss it with the person who implemented and to be sure with them. 
- For the next sprint, the compressor can be implemented for profile pictures also.

### DIFF
--- a/app/src/androidTest/java/com/android/ootd/ui/post/AddItemsViewModelFirebaseTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/ui/post/AddItemsViewModelFirebaseTest.kt
@@ -306,7 +306,7 @@ class AddItemsViewModelFirebaseTest : FirestoreTest() {
     viewModel.setStyle("Minimalist")
     viewModel.setNotes("Worn twice")
 
-    viewModel.onAddItemClick()
+    viewModel.onAddItemClick(context)
     kotlinx.coroutines.delay(2000)
 
     assertTrue(viewModel.addOnSuccess.first())


app/src/main/java/com/android/ootd/ui/post/items/AddItemsScreen.kt
```diff
@@ -155,7 +155,7 @@ fun AddItemsScreen(
 
                 // image pick
                 showDialog = showDialog,
-                onShowDialogChange = { showDialog = it },
+                onShowDialogChange = { newValue -> showDialog = newValue },
```
Clemensito: I don't understand the change here. Was this necessary ? I am quite confused 😅

app/src/main/java/com/android/ootd/ui/post/items/AddItemsScreen.kt
```diff
@@ -183,15 +183,15 @@ fun AddItemsScreen(
                 brand = itemsUIState.brand,
                 onBrandChange = addItemsViewModel::setBrand,
                 price = itemsUIState.price,
-                onPriceChange = addItemsViewModel::setPrice,
+                onPriceChange = { addItemsViewModel.setPrice(it) },
```
Clemensito: Same here, I am not quite sur to understand. Both lines do the same, no ?

Clemensito (commented):
Good PR overall. I would need you to take a look at my comments but other than that I think we are good to go ! Really good job 🔥

app/src/main/java/com/android/ootd/model/items/FirebaseImageUploader.kt
```diff
@@ -15,7 +15,9 @@ import kotlinx.coroutines.tasks.await
 object FirebaseImageUploader {
 
   private const val TAG = "FirebaseImageUploader"
-  private const val UPLOAD_TIMEOUT_MS = 1_000L
+
+  // Timeout duration for upload operations to Firebase Storage
+  private const val UPLOAD_TIMEOUT_MS = 10_000L // 10 seconds
```
cocoStein: I understand that you changed this to a 10s timer so if you try to upload a large picture with bad internet it can still try? But now, the fallback of the upload to Firebase not succeeding is caching the element locally and then only uploading it once back online. 

So if you are offline, you need to wait 10s until you are sent back to the `inventory` or to `post preview` screen. One solution could be to by default store it locally and then to upload it to the firebase afterward. But this requires more changes of the current setup, so for now i think that we should put this back to a lower value.

app/src/main/java/com/android/ootd/ui/post/items/AddItemsScreen.kt
asunluer: Yes you are right ! Good catch !

app/src/main/java/com/android/ootd/model/items/FirebaseImageUploader.kt
cocoStein: What I want to add, which I think partially works already. Is to make some repository "Offline first“. This means that for example when you add a new item it's first added locally (so for the user it's seamless and he can do whatever he wants now) and in the background, if he's connected to the internet, the item is sent. These Firebase instructions will be saved in a queue so if for example you `add a new item -> Edit it -> delete it` then the item will not show in the Firebase !

Hope this helps, feel free to ask if you have any questions of if anything i said wasn't clear 🔥

asunluer:
Thank you guys so much ! I am appreciated with your reviews. It was nice to talking with you :)

PR #385, opened 2025-11-22
Fix: Inventory loading & Search Bar

cocoStein:
# Summary
In this PR, I changed the logic of showing the item list in the inventory to show the cached list if it exists. This makes it so that you don't have to constantly reload the list each time you come to the screen. I added a fallback to update the list in the background, but as the item list can only be changed locally, this should never be used, except if the user had the same account on 2 different devices.

I also added a search function on the inventory screen.

# What

- UI
  - used the new custom Composable for the  `OOTDTopBar`
  - Created a custom composable for the `InventorySearchBar`
  - Added a new `FloatingActionButton` that toggles the search bar
 
- ViewModel
  - Added the search logic, you can filter by `type`, `category`, `style`, `notes`
>[!NOTE]
> I asked some old tester what field they would want to test. These are the ones that came up but it's not a final list. We can easily change this !
  - Changed the display logic to first check if a local cached version of the item list exist, if it's the case then we display that and make the firebase calls in the background.
- Tests
  - Unit tests added/updated (describe what and where)
  - Integration tests added/updated
  - Removed old tests as they are not useful anymore with the new cached lists

# Why
If the user has a lot of items, then you have to scroll through the whole list. This is bad UX.
The constant reloading was painful, as it would make the app appear slow, and if you had bad internet, it could ruin the process.

# Screenshots / Videos
This video shows the new cached list. Now you don't have a loading screen each time you navigate to the screen

[Screen_recording_20251122_143044.webm](https://github.com/user-attachments/assets/0745ab58-9452-4fbc-ab92-20248511c2ad)

This video shows the flow of searching items.

[Screen_recording_20251122_143135.webm](https://github.com/user-attachments/assets/d31d8ba2-9427-439c-aee3-7906698917b4)

# Checklist
- [x] Manual verification: app run and basic flows checked (list steps in

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/Inventory/InventoryScreen.kt
+++ b/app/src/main/java/com/android/ootd/ui/Inventory/InventoryScreen.kt
@@ -27,6 +27,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Color.Companion.Red
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
@@ -119,7 +120,7 @@ fun InventoryScreen(
 
           FloatingActionButton(
               onClick = { inventoryViewModel.toggleSearch() },
-              containerColor = Primary,
+              containerColor = if (uiState.isSearchActive) Red else Primary,
               modifier = Modifier.testTag(InventoryScreenTestTags.SEARCH_FAB)) {
                 Icon(
                     imageVector =
--- a/app/src/main/java/com/android/ootd/ui/Inventory/InventoryViewModel.kt
+++ b/app/src/main/java/com/android/ootd/ui/Inventory/InventoryViewModel.kt
@@ -81,7 +81,13 @@ class InventoryViewModel(
             // Display cached items immediately
             val sortedCachedItems = sortItemsByCategory(cachedItems)
             allItems = sortedCachedItems
-            _uiState.value = _uiState.value.copy(items = sortedCachedItems, isLoading = false)
+            val displayItems =
+                if (_uiState.value.isSearchActive) {
+                  filterItems(sortedCachedItems, _uiState.value.searchQuery)
+                } else {
+                  sortedCachedItems
+                }
+            _uiState.value = _uiState.value.copy(items = displayItems, isLoading = false)
           }
         }
 
--- a/app/src/test/java/com/android/ootd/ui/Inventory/InventoryViewModelTest.kt
+++ b/app/src/test/java/com/android/ootd/ui/Inventory/InventoryViewModelTest.kt
@@ -280,4 +280,32 @@ class InventoryViewModelTest {
 
     Assert.assertTrue(viewModel.uiState.value.items.isEmpty())
   }
+
+  @Test
+  fun `background refresh updates items if different`() = runTest {
+    val newItems = listOf(testItems[0])
+    coEvery { accountRepository.getItemsList(testUserId) } returnsMany
+        listOf(testItemIds, listOf("item1"))
+    coEvery { itemsRepository.getItemsByIds(testItemIds) } returns testItems
+    coEvery { itemsRepository.getItemsByIds(listOf("item1")) } returns newItems
+
+    viewModel = InventoryViewModel(accountRepository, itemsRepository)
+    advanceUntilIdle()
+
+    Assert.assertEquals(newItems, viewModel.uiState.value.items)
+  }
+
+  @Test
+  fun `background refresh failure is silent if cache exists`() = runTest {
+    coEvery { accountRepository.getItemsList(testUserId) } returns
+        testItemIds andThenThrows
+        Exception("Network")
+    coEvery { itemsRepository.getItemsByIds(testItemIds) } returns testItems
+
+    viewModel = InventoryViewModel(accountRepository, itemsRepository)
+    advanceUntilIdle()
+
+    Assert.assertEquals(testItems, viewModel.uiState.value.items)
+    Assert.assertNull(viewModel.uiState.value.errorMessage)
+  }
 }


app/src/main/java/com/android/ootd/ui/Inventory/InventoryViewModel.kt
```diff
@@ -56,31 +63,61 @@ class InventoryViewModel(
...
    * Loads the user's inventory items using offline-first pattern.
    *
-   * Uses short timeout (5 seconds) for local cache reads. With Firestore persistence, reads from
-   * cache complete quickly even offline.
+   * First loads from cache immediately (non-blocking), then fetches fresh data in the background.
+   * This provides instant UI updates from cache while ensuring data freshness.
    */
   fun loadInventory() {
     viewModelScope.launch {
-      _uiState.value = _uiState.value.copy(isLoading = true, errorMessage = null)
```
asunluer: I would like to ask you why did you erase `isLoading = true` ?

app/src/main/java/com/android/ootd/ui/Inventory/InventoryScreen.kt
```diff
@@ -91,57 +98,82 @@ fun InventoryScreen(
   Scaffold(
       modifier = modifier.fillMaxSize().testTag(InventoryScreenTestTags.SCREEN),
       topBar = {
-        CenterAlignedTopAppBar(
```
asunluer: Good design choice by removing it and use our theme ! ;)

app/src/main/java/com/android/ootd/ui/Inventory/InventoryScreen.kt
```diff
@@ -91,57 +98,82 @@ fun InventoryScreen(
...
+                Icon(
+                    imageVector =
+                        if (uiState.isSearchActive) Icons.Default.Close else Icons.Default.Search,
+                    contentDescription =
+                        if (uiState.isSearchActive) "Close Search" else "Search Item",
+                    tint = MaterialTheme.colorScheme.onPrimary)
+              }
+        }
       },
       snackbarHost = { SnackbarHost(hostState = snackbarHostState) }) { innerPadding ->
```
asunluer: I don't understand why you used snackbarHost ? What is your intuition or idea in here because I didn't see any UI design related to that ? Or, what do I need to do that to see ?

app/src/main/java/com/android/ootd/ui/Inventory/InventoryViewModel.kt
```diff
@@ -56,31 +63,61 @@ class InventoryViewModel(
...
+                    sortedFreshItems
+                  }
+              _uiState.value = _uiState.value.copy(items = displayItems, isLoading = false)
+            }
+          } catch (e: Exception) {
+            // Silently fail background refresh -> user already sees cached data
+            if (_uiState.value.items.isEmpty()) {
+              _uiState.value =
+                  _uiState.value.copy(
+                      isLoading = false, errorMessage = "Failed to load inventory: ${e.message}")
```
asunluer: I wonder is there a reason why you did not want to show an error message to users when the loading fails for inventory ?

asunluer (changes requested):
I looked at your implementation and very good code and very well explained codes. Thank you so much for this well explained PR. 

Also, I liked the flow of the search bar. It is very smooth, well designed and searching is so easily. You did a very good job on that and thank you so much for your time ;). However, I would like to mention a part from a user eye, when I finished to search the search bar stays even I tried to close it by clicking a random places in the screen in the screen and I did it by intuition, like I did not realised at first that I had to click on the FAB, we can keep the closing FAB but also it can be better to close the search bar by clicking a random part on the screen. What do you think about it ? 

We can do tests also to see the users reaction without telling how to close a search bar. Let me know about this part. 

Overall, it is really good !! I liked a lot :)

Note: 
* Do not forget to increase the coverage pourcentage and let me know if you need help for this. 
* There were 2 issues on Sonar Cloud side, when I go to the web page, it shows nothing maybe you handled them already.

app/src/main/java/com/android/ootd/ui/Inventory/InventoryViewModel.kt
cocoStein: A personal choice, as personally I don't feel like users enjoy seeing error messages. If you suggest me to add one, I can do it !

app/src/main/java/com/android/ootd/ui/Inventory/InventoryViewModel.kt
```diff
@@ -56,31 +63,61 @@ class InventoryViewModel(
...
-        // Fetch all items using the batch method (with timeout for offline)
-        val items =
-            kotlinx.coroutines.withTimeoutOrNull(FIRESTORE_TIMEOUT_MS) {
-              itemsRepository.getItemsByIds(itemIds)
-            } ?: emptyList()
+          if (cachedItems.isNotEmpty()) {
+            // Display cached items immediately
+            val sortedCachedItems = sortItemsByCategory(cachedItems)
+            allItems = sortedCachedItems
+            _uiState.value = _uiState.value.copy(items = sortedCachedItems, isLoading = false)
```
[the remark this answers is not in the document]
cocoStein: Good catch ! My emulator must be to quick because i never saw this !!

Thanks for the suggestion, i'll try it on my end but from what i see it look to fix it !

app/src/main/java/com/android/ootd/ui/Inventory/InventoryScreen.kt
cocoStein: This is triggered by (around line 90):

```Kotlin
  // Show error message if present
  LaunchedEffect(uiState.errorMessage) {
    uiState.errorMessage?.let { message ->
      snackbarHostState.showSnackbar(message) // This line triggers the snackbat
      inventoryViewModel.clearError()
    }
  }
```

This is triggered if the `InventoryViewModel ` catches any exception. You can see it by creating an error, but it's good news if you don't see it 😄. 

Let me know if you have any other questions !

cocoStein:
Hey, thanks for the feedback. I'll fix the glitchy screen when you get back to the inventory screen !

For getting out of focus of the search screen, I didn't know if when you are out of focus it also means that you want to stop searching (you get back to the normal inventory), or you want to scroll through the items you filtered just without the search bar. I don't know if you see what I mean.
One thing I can do is to make the `Close search bar` button red, so a new user can see it more. I totally agree with you that it's not very obvious at first.

Let me know what you think ! 🔥

app/src/main/java/com/android/ootd/ui/Inventory/InventoryViewModel.kt
asunluer: Yes, I think I get it you mean that since the items are cached you do not need to show loading screen because they will be already loaded, if the cached items are empty then it will load the page. Yes I see you point. Thank you :)

PR #393, opened 2025-11-24
Feat/382 change design for notifications for follow request

stefantaga24:
# Summary
Added accept and delete buttons for notification. Now the notification also contains the username of the sender.

# What

- UI
  - No UI was changed except the MainActivity where I had to change the way the notification was built.
- ViewModel
  - Notifications now have the senderName as a parameter in the data class in order to build the correct notification message.
  - We now have a NotificationReceiver which has the job to get the actions of the user in the notifications and then modify the database accordingly. Therefore, if the user presses accept, we accept the follow notification.
- Tests
  - Updated tests to handle the new senderName parameters.
  - Added tests for NotificationReceiver.

# Why

With this, notifications can be accepted directly from the bar, users do not need to go to the notifications screen anymore. This will tie nicely when we can send notifications to users even when they are not logged in into the app.

# Screenshots / Videos

This is how the notification should look:

image

https://github.com/user-attachments/assets/f650dbfd-7c7a-467a-92e3-16fb712b2deb

### DIFF
--- a/app/src/androidTest/java/com/android/ootd/model/notifications/NotificationReceiverTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/model/notifications/NotificationReceiverTest.kt
@@ -54,7 +54,7 @@ class NotificationActionReceiverInstrumentedTest : FirestoreTest() {
   }
 
   @Test
-  fun testAcceptActionTriggersRepositoryCall() = runTest {
+  fun testAcceptAction() = runTest {
     val notif =
         Notification(
             uid = "notif123",
@@ -83,5 +83,37 @@ class NotificationActionReceiverInstrumentedTest : FirestoreTest() {
     assert(result.isEmpty())
 
     verify { mockManager.cancel(notif.uid.hashCode()) }
+  }
+
+  @Test
+  fun testDeleteAction() = runTest {
+    val notif =
+        Notification(
+            uid = "notif123",
+            senderId = currentUser.uid,
+            receiverId = currentUser.uid,
+            type = "FOLLOW_REQUEST",
+            content = "hello",
+            senderName = "")
+
+    notificationsRepository.addNotification(notif)
+
+    val intent =
+        Intent().apply {
+          action = NOTIFICATION_ACTION_DELETE
+          putExtra("notificationUid", notif.uid)
+          putExtra("senderId", notif.senderId)
+          putExtra("receiverId", notif.receiverId)
+        }
+
+    receiver.onReceive(mockContext, intent)
+
+    // Advance time to allow coroutines to complete
+    advanceUntilIdle()
+    val result = notificationsRepository.getNotificationsForReceiver(currentUser.uid)
+    assert(result.isEmpty())
+    verify {
+      mockManager.cancel(notif.uid.hashCode())
+    } // Make sure that the notification was canceled
   }
 }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/ootd/model/account/AccountRepositoryProvider.kt
```diff
@@ -8,9 +8,10 @@ import com.google.firebase.firestore.firestore
...
 object AccountRepositoryProvider {
-  private val _repository: AccountRepository by lazy {
-    AccountRepositoryFirestore(Firebase.firestore)
+
+  private fun defaultRepository(): AccountRepository {
+    return AccountRepositoryFirestore(Firebase.firestore)
   }
 
-  var repository: AccountRepository = _repository
+  var repository: AccountRepository = defaultRepository()
```
cocoStein: I don't understand why you made this change for the Account repository. The only thing that you changed is now it's not `lazy` ? Isn't that supposed to create a problem with the Firebase functions ? (I tried to interact with the app and everything seems to be in order, so I don't know what this change is.)

app/src/main/java/com/android/ootd/model/notifications/NotificationRepositoryFirestore.kt
```diff
@@ -205,4 +209,22 @@ class NotificationRepositoryFirestore(private val db: FirebaseFirestore) : Notif
...
+
+    // Add the sender as a friend
+    val wasAddedToBoth = accountRepository.addFriend(receiverId, senderId)
+
+    // If I could not update both friend lists
+    // I throw an exception such that the notification does not disappear.
+    // This will also help with offline mode
+    check(wasAddedToBoth) { "Could not update both friend lists" }
+    // Delete the notification
+    deleteNotification(notificationId = notificationId, receiverId = receiverId)
```
cocoStein: Here, why do you say that you throw an error "Such that the notification does not disappear" but then on line 228 you `deleteNotification` ?

I like the idea of keeping the notification if the interaction did not work, but sometimes it can create some false positives. Such as here where I asked 2 accounts to follow each other (both got notifications) and I accepted it on one of them. But the notification is still present in the other one, and your fallback for this is to `retry`. But I think in this case, we should maybe have a fallback to see if the other person is not already your friend.

Let me know what you think of this fallback ?

https://github.com/user-attachments/assets/cf479900-273a-41e1-8472-db986e8c1699

app/src/main/java/com/android/ootd/MainActivity.kt
```diff
@@ -204,27 +207,67 @@ fun OOTDApp(
...
+        PendingIntent.getBroadcast(
+            context,
+            ("${notification.uid}_delete").hashCode(),
+            deleteIntent,
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE)
+
+    // BUILD NOTIFICATION
     val builder =
-        NotificationCompat.Builder(context, "ootd_channel")
+        NotificationCompat.Builder(context, OOTD_CHANNEL_ID)
```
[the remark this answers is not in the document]
stefantaga24: Thanksss

app/src/main/java/com/android/ootd/model/account/AccountRepositoryProvider.kt
cocoStein: Ok perfect 👍🏼

app/src/main/java/com/android/ootd/model/notifications/NotificationRepositoryFirestore.kt
cocoStein: No worries this is why we do code review 🔥  The rest of your code was already very impressive, I have to think hard to find edge cases !

Issue #394, opened 2025-11-24
Get notifications even when the user does not have the app opened


stefantaga24:
I will implement a listener at the android level such that we are able to send notifications to the user even when they do not have the app opened.

PR #427, opened 2025-11-27
Feat/394 get notifications even when the user does not have the app opened

stefantaga24:
# Summary

Added background worker to check for notifications when the user is not on the app.

# What

- ViewModel
  - A background worker is started when the main activity is created.
  - Created new coroutines for getting unpushed notifications needed for the background worker.
- Tests
  - Added tests for the background worker
  - Added tests for the new repository functions used by the background worker
  - Added a test for the main activity to simulate a user clicking a notification leading to a user

# Why

This helps with issue #394 such that a user does not have to open the app in order to get notifications.

### DIFF
--- a/app/src/main/java/com/android/ootd/model/notifications/NotificationRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/ootd/model/notifications/NotificationRepositoryFirestore.kt
@@ -14,6 +14,7 @@ import kotlinx.coroutines.tasks.await
 const val NOTIFICATION_COLLECTION_PATH = "notifications"
 const val SENDER_ID = "senderId"
 const val RECEIVER_ID = "receiverId"
+const val TAG = "NotificationRepositoryFirestore"
 
 @Keep
 data class NotificationDto(
@@ -48,6 +49,8 @@ private fun NotificationDto.toDomain(): Notification {
       senderName = this.senderName)
 }
 
+private const val WAS_PUSHED = "wasPushed"
+
 class NotificationRepositoryFirestore(private val db: FirebaseFirestore) : NotificationRepository {
 
   /** Helper method to check notification data as firestore might add the default values */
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/android/ootd/model/notifications/NotificationRepositoryFirestore.kt
```diff
@@ -97,6 +97,39 @@ class NotificationRepositoryFirestore(private val db: FirebaseFirestore) : Notif
...
+          db.collection(NOTIFICATION_COLLECTION_PATH)
+              .whereEqualTo(RECEIVER_ID, receiverId)
+              .whereEqualTo("wasPushed", false)
+              .get()
+              .await()
+
+      querySnapshot.documents.mapNotNull { document -> transformNotificationDocument(document) }
+    } catch (e: Exception) {
+      Log.e(
+          "NotificationRepositoryFirestore",
```
cocoStein: nitpicky, but if you want to keep this log and the one in l126, I think it's best if you create a companion object with a TAG, so we follow the Google coding standard. !

app/src/main/java/com/android/ootd/model/notifications/NotificationsUtil.kt
```diff
@@ -0,0 +1,106 @@
...
+              R.drawable.ic_delete, // delete icon
+              "Delete",
+              deletePendingIntent)
+
+  manager.notify(notification.uid.hashCode(), builder.build())
+}
+
+fun scheduleBackgroundNotificationSync(context: Context) {
+  val workRequest =
+      PeriodicWorkRequestBuilder<NotificationSyncWorker>(15, TimeUnit.MINUTES)
```
cocoStein: Is there a technical reason why you chose 15 min ?

PR #406, opened 2025-11-25
Feature : Wishlist V1 + tighter items fields

MarcK0909:
## Summary
Added the wishlist (“Starred”) feature end to end and tightened the Add/Edit Item form experience, including enforcing character limits (to clean up the issue mentioned in #379 ), cleaning up previews, and expanding test coverage. The wishlist so far lets users star their favorite items from their inventory to see them in a new tab in their account. This is of course a work in progress and will be further refined/updated as we see fit.

### UI
- Inventory grid, Account page, and Select-from-inventory screens now show **star toggles** (yellow icon).
- Account page includes a **“Starred” tab** with its own grid.
- See Fit dialog layout refined (bold labels, copy buttons, scrollable preview).
- Edit Item buttons exposed safely in tests.
- Added **live character counters/limits** across item fields.
- Tweaked See Fit dialog spacing.
- Star icons positioned in the top-right of cards.
- Introduced **StarYellow** color for the star icon.

### ViewModel
- `AccountRepository`, Firestore/InMemory implementations, and `AccountPageViewModel` gained starred-item state, fetching, and toggle flows.
- `InventoryViewModel` now tracks starred IDs and exposes `toggleStar`.
- Select-from-inventory view continues to pass no-op star callbacks so it stays read-only.

### Documentation
- Added kdoc where relevant for new functions.

### Tests
- **Unit tests:** star toggling (InventoryViewModel, AccountPageViewModel, repo tests), focus-only counters (ItemFormSharedComponentsTest), override-photo flow (AddItemsViewModelTest).
- **Instrumented tests:** Edit Item scroll-safe assertions, Inventory screen verifies star icon visibility, Select Inventory fake repo updated for toggle API.

## Why
We want users to build a wishlist of inventory pieces they love, so we had to add end-to-end support for “starred” items—from persisting the IDs on the account, syncing that state in the Inventory/Account view models, to showing a dedicated tab and star toggles in the UI. At the same time, feedback on the item forms called for harder limits (brand/material/notes, etc.) so the preview and See Fit dialog stay clean and consistent; enforcing those caps centrally keeps Add/Edit flows aligned. Finally, the new unit/UI tests lock down those behaviors so future tweaks to the shared form components or star logic can’t silently regress the feature.

## Screenshots / Videos
New "Starred" tab and star toggle (I just changed the color since the video was taken but nothing else) :

https://github.com/user-attachments/assets/05a8b60a-27c9-4040-82f8-64a34e0deb45

Focus-only character counters on item fields : 

[...]

PR #436, opened 2025-11-28
Feat : Wishlist v2

MarcK0909:
## Summary
Implemented “Wishlist v2” so users can star friends’ items from the See Fit feed, have those selections sync into their account wishlist, and hardened the flaky item form tests that were failing on CI. This ties in to the first version of this feature (#406) and enables more user interaction as discussed with PO. We opted to reference the item and not copy it directly as this would pose data ownsership issues regarding the items, we felt a reference struck the correct balance for the user to have a good experience while not "stealing" from other users. 

>[!WARNING]
The PR slightly exceeds the 1000 line threshold but most of it is just tests I had to add due to the coverage rules

### UI
- Added a **star toggle** to See Fit item cards for non-owners plus a details dialog entry point from the wishlist grid so users can inspect starred pieces without leaving their profile.
- Account tab now **refreshes wishlist data** when the user returns to the screen and keeps the starred grid in sync with the latest IDs, showing a dedicated **empty state** until something is starred.

### ViewModel / Data
- `SeeFitViewModel` now talks to `AccountRepository` to load starred IDs, toggle stars, and expose that state to the grid.
- `AccountPageViewModel` refreshes starred IDs (including friend-owned items) via the new **cross-owner item lookup** and keeps optimistic UI state while Firestore catches up.
- Items/Account repositories gained:
  - `getItemsByIdsAcrossOwners`
  - `refreshStarredItems`
- Firestore implementation batches reads via **document-id queries**.
- Local/fake repositories updated accordingly.
- Account screen hooks into lifecycle events to **eagerly refresh** or warn.

### Tests
- Added **Firebase-emulator coverage** for `ItemsRepositoryFirestore` and `ItemsRepositoryLocal` along with instrumentation for the new See Fit, preview, and select-inventory flows exercising starred-item handling.
- Unit tests now cover the new See Fit VM + Account Page VM branches.
- Stabilized flaky `AddItemScreenTest` material field interactions by driving the ViewModel directly and waiting for the additional-details section before scrolling.

## Why
Give users a consistent wishlist story: they can now star pieces they like in friends’ outfits, see those items instantly inside their profile, and the repositories/tests guarantee the data layer behaves (even when items belong to other owners or when offline).

## Screenshots / Videos
How the star presents itself when fiewing a user's Fit : 

Screenshot_20251128_142546

The whole process of starring an item and then unstarring it : 

https://github.com/user-attachments/assets/4955eb95-3a5e-45a8-860d-4ec642360f1e

## Checklist
- [x] Manual verification: app run and basic flows checked, starring an item adds it to the tab in Account

### DIFF
--- a/app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
@@ -12,6 +12,7 @@ const val ITEMS_COLLECTION = "items"
 const val OWNER_ATTRIBUTE_NAME = "ownerId"
 const val POST_ATTRIBUTE_NAME = "postUuids"
 const val NOT_LOGGED_IN_EXCEPTION = "ItemsRepositoryFirestore: User not logged in."
+private const val TAG = "ItemsRepositoryFirestore"
 
 class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsRepository {
 
@@ -87,14 +88,12 @@ class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsReposit
 
     val cachedItems = uuids.mapNotNull { itemsCache[it] }
     if (cachedItems.size == uuids.size) {
-      Log.d("ItemsRepositoryFirestore", "Returning all ${uuids.size} items from memory cache")
+      Log.d(TAG, "Returning all ${uuids.size} items from memory cache")
       return cachedItems
     }
 
     if (cachedItems.isNotEmpty()) {
-      Log.d(
-          "ItemsRepositoryFirestore",
-          "Found ${cachedItems.size}/${uuids.size} items in cache, fetching rest")
+      Log.d(TAG, "Found ${cachedItems.size}/${uuids.size} items in cache, fetching rest")
     }
 
     val missingUuids = uuids.filterNot { itemsCache.containsKey(it) }
@@ -113,11 +112,9 @@ class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsReposit
       kotlinx.coroutines.withTimeoutOrNull(2_000L) {
         missingUuids.chunked(10).forEach { batch -> fetched += fetchBatch(batch, ownerFilter) }
       }
-      Log.d(
-          "ItemsRepositoryFirestore",
-          "Fetched ${fetched.size} items from Firestore (missing=${missingUuids.size})")
+      Log.d(TAG, "Fetched ${fetched.size} items from Firestore (missing=${missingUuids.size})")
     } catch (e: Exception) {
-      Log.w("ItemsRepositoryFirestore", "Failed to fetch missing items (offline): ${e.message}")
+      Log.w(TAG, "Failed to fetch missing items (offline): ${e.message}")
     }
     return fetched
   }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
```diff
@@ -658,6 +653,35 @@ class AccountRepositoryFirestore(
...
+    val cacheResult = fetchItemsFromSource(userId, Source.CACHE)
+    if (!cacheResult.isNullOrEmpty()) return cacheResult
+    val serverResult = fetchItemsFromSource(userId, Source.SERVER)
+    return serverResult ?: cacheResult ?: emptyList()
+  }
+
+  private suspend fun fetchItemsFromSource(userId: String, source: Source): List<String>? {
+    val timeout = if (source == Source.CACHE) 1_000L else 2_000L
+    return try {
+      withTimeoutOrNull(timeout) {
```
bbianca2004: The refactor in this file looks great! Extracting Firestore read logic into `loadExistingItems()` and `fetchItemsFromSource()` removes duplication and centralizes timeout and fallback handling. Altogether this makes the repository more maintainable and ensures more reliable data flow across the app.

app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
```diff
@@ -57,64 +58,96 @@ class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsReposit
...
+        missingUuids.chunked(10).forEach { batch -> fetched += fetchBatch(batch, ownerFilter) }
       }
+      Log.d(
+          "ItemsRepositoryFirestore",
+          "Fetched ${fetched.size} items from Firestore (missing=${missingUuids.size})")
+    } catch (e: Exception) {
+      Log.w("ItemsRepositoryFirestore", "Failed to fetch missing items (offline): ${e.message}")
     }
+    return fetched
+  }
```
bbianca2004: Nitpick: we use "ItemsRepositoryFirestore" inline in more of places here. Might be nice to define a single private const val TAG for consistency.

app/src/main/java/com/android/ootd/ui/account/AccountScreenViewModel.kt
```diff
@@ -132,15 +136,33 @@ class AccountPageViewModel(
...
+                  if (updatedIds.contains(item.itemUuid)) {
+                    if (state.starredItems.any { it.itemUuid == item.itemUuid }) {
+                      state.starredItems
+                    } else {
+                      state.starredItems + item
+                    }
+                  } else {
+                    state.starredItems
+                  })
+        }
```
bbianca2004: Noticed here in toggleStar(), the optimistic update only adds items when starring but doesn't remove them when unstarring, so items stay visible in the wishlist until the screen refreshes(only when switching screens). It would be nice to remove the item from starredItems immediately so the UI reflects the change without waiting for a lifecycle refresh. Cause I notice that the unscarred items only disappear from the wishlist by a screen refresh and I was wondering if maybe it could happen instantaneously? Let me know if this is feasible or not, I am sure you know best :))

app/src/test/java/com/android/ootd/model/items/ItemsRepositoryFirestoreTest.kt
```diff
@@ -0,0 +1,88 @@
...
+  @After
+  fun tearDown() {
+    unmockkAll()
+  }
+
+  @Test
+  fun `fetchCrossOwnerBatch caches fetched items`() = runTest {
+    val ids = listOf("item-1", "item-2")
+    val item1 = sampleItem("item-1")
+    val item2 = sampleItem("item-2")
```
bbianca2004: Nitpick: I feel like this test and `getItemsByIdsReturnsCachedItemsAfterDeletion` in the emulated tests both verify that `ItemsRepositoryFirestore` returns cached items instead of refetching them. The emulator test already validates this behavior end-to-end, so the mock-based one might be partially redundant. Not a big deal though but maybe you could consider removing the emulated one since it is slower on the ci. Let me know what you think though and if I missed something :))

app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
MarcK0909: Yes good idea

app/src/main/java/com/android/ootd/ui/account/AccountScreenViewModel.kt
bbianca2004: Thanks for your explanation! It makes sense what you say, it would be annoying if you accidentally unstar an item. I was just thinking that a future improvement would be to consider using a Snackbar with some sort of Undo action when unstarring or maybe a confirmation pop up for when unstarring an item, I think this would also make the page fully responsive while still giving the user a change to undo if he makes a mistake. But this is not a very big issue at the moment, what I said could maybe be done as future improvement.

bbianca2004 (approved):
Thank you again for your work on this feature, it looks great🫡 The implementation is great, the UI is very user friendly and the tests are very complete, well done!! Also thanks for taking into my account my suggestions it all looks good to me now, you can safely merge :))

PR #439, opened 2025-11-29
Fix: Offline mode for inventory

cocoStein:
# Summary
Refractor the offline mode, so it works seamlessly. Now if you create new item while being offline. Once you get back online they are uploaded to the firebase. 

>[!NOTE]
> If you added many items offline. Once you get back online, the image can take a bit of time to be uploaded if you have bad internet !

# What
Native Firestore Persistence:
Refactored `ItemsRepositoryFirestore.kt` and `AccountRepositoryFirestore.kt` to remove all manual in-memory caching (itemsCache, itemsListCache, etc.) and `withTimeoutOrNull` blocks.

The app now relies on the Firestore SDK's built-in offline persistence, which automatically queues writes and syncs them when online, preventing data loss and race conditions.

Background Image Uploads:
Created a new `ImageUploadWorker` using Android's WorkManager.
This worker handles uploading images to Firebase Storage when the device regains connectivity and updates the corresponding Firestore document with the correct public URL.

ViewModel Integration:
Updated `AddItemsViewModel.kt` and `EditItemsViewModel.kt` to detect when an image upload returns a local URI (indicating offline mode).
In these cases, the `ViewModels` now enqueue a `OneTimeWorkRequest` for the `ImageUploadWorker`.
The item is saved to Firestore immediately (queued locally), allowing the user to see their new item instantly, while the background worker ensures the image is properly synced to the server later.

Uploader Logic:
Enhanced `FirebaseImageUploader.kt` with a `uploadImageSuspending` method designed for the background worker, which throws exceptions instead of suppressing them, allowing WorkManager to retry failed uploads automatically.

Updated some tests that caused merge conflicts and added new one to test this functionnality

# Why
A requirement for milestone 3 is to have offline capabilities in the app. This also improves the user experience as they don't have to be constantly online to use the app.

# Screenshots / Videos

Video showing the full flow of adding new item while being offline and then going back online.

https://github.com/user-attachments/assets/1b3e83cd-17e9-4f32-a5dc-962f792e002f

# Checklist

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/post/items/AddItemsViewModel.kt
+++ b/app/src/main/java/com/android/ootd/ui/post/items/AddItemsViewModel.kt
@@ -219,7 +219,6 @@ open class AddItemsViewModel(
       // Cache updates happen synchronously before Firestore .await()
       try {
         repository.addItem(item)
-        Log.d(TAG, "Item added to cache, Firestore queued")
       } catch (e: Exception) {
         // Acceptable when offline - cache is still updated
         Log.w(TAG, "Item add may be offline (cache updated): ${e.message}")
@@ -227,14 +226,12 @@ open class AddItemsViewModel(
 
       try {
         accountRepository.addItem(item.itemUuid)
-        Log.d(TAG, "Account updated in cache, Firestore queued")
       } catch (e: Exception) {
         // Acceptable when offline - cache is still updated
         Log.w(TAG, "Account add may be offline (cache updated): ${e.message}")
       }
 
       // Operations completed - cache is updated, Firestore will sync
-      Log.d(TAG, "Item operations completed (cache updated)")
       true
     } catch (e: Exception) {
       Log.e(TAG, "Error in item operations: ${e.message}", e)
@@ -288,7 +285,6 @@ open class AddItemsViewModel(
                         .build()
 
                 WorkManager.getInstance(context).enqueue(workRequest)
-                Log.d(TAG, "Scheduled background upload for item $itemUuid")
               }
 
               uploaded
--- a/app/src/main/java/com/android/ootd/ui/post/items/EditItemsViewModel.kt
+++ b/app/src/main/java/com/android/ootd/ui/post/items/EditItemsViewModel.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.net.Uri
 import android.util.Log
 import android.webkit.URLUtil.isValidUrl
+import androidx.core.net.toUri
 import androidx.lifecycle.viewModelScope
 import androidx.work.Constraints
 import androidx.work.NetworkType
@@ -185,7 +186,7 @@ open class EditItemsViewModel(
                       compressedImage, state.itemId, state.localPhotoUri)
 
               // Check if we need to schedule background upload
-              val uri = Uri.parse(uploaded.imageUrl)
+              val uri = uploaded.imageUrl.toUri()
               if (uri.scheme == "content" || uri.scheme == "file") {
                 val workRequest =
                     OneTimeWorkRequestBuilder<ImageUploadWorker>()
@@ -201,7 +202,6 @@ open class EditItemsViewModel(
                         .build()
 
                 WorkManager.getInstance(context).enqueue(workRequest)
-                Log.d(TAG, "Scheduled background upload for item ${state.itemId}")
               }
               uploaded
             } else state.image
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
```diff
@@ -436,75 +428,33 @@ class AccountRepositoryFirestore(
...
-        }
-        Log.d(TAG, "Item added to Firestore (or queued if offline)")
-      } catch (e: Exception) {
-        // Acceptable when offline - cache is already updated
-        Log.w(TAG, "Firestore update queued (offline): ${e.message}")
-      }
-
-      true // Cache is updated, that's what matters
+      withTimeout(2000L) { userRef.update("itemsUids", FieldValue.arrayUnion(itemUid)).await() }
+      Log.d(TAG, "Item added to Firestore")
```
asunluer: If you use this line to debug, it is good that you remove it. Otherwise, if you want to keep a log for this and since you do not use any specific value to indicate this message, you can use `Log.i()`. Let me know, what you think about this ?

app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
```diff
@@ -608,78 +547,39 @@ class AccountRepositoryFirestore(
...
-          document?.getStringList("starredItemUids") ?: emptyList()
+          withTimeout(1000L) { userRef.get().await() }
         } catch (e: Exception) {
-          Log.w(TAG, "Could not fetch starred items: ${e.message}")
-          emptyList()
+          // If read fails/times out, we can't toggle reliably without cache.
+          // But we can try to read from cache specifically
+          try {
+            userRef.get(Source.CACHE).await()
+          } catch (e2: Exception) {
```
asunluer: This exception e2 is not used you can remove it to `_`.

app/src/main/java/com/android/ootd/model/items/ItemsRepositoryFirestore.kt
```diff
@@ -76,141 +99,51 @@ class ItemsRepositoryFirestore(private val db: FirebaseFirestore) : ItemsReposit
...
 
-    // Queue Firestore update with timeout (will sync when online, doesn't block)
-    try {
-      kotlinx.coroutines.withTimeoutOrNull(2_000L) {
-        db.collection(ITEMS_COLLECTION).document(itemUUID).set(newItem).await()
-      }
-      Log.d(TAG, "Item edited in Firestore (or queued if offline)")
+      withTimeout(2000L) { db.collection(ITEMS_COLLECTION).document(itemUUID).set(newItem).await() }
+    } catch (e: TimeoutCancellationException) {
+      // Timeout means offline; Firestore will sync later
```
asunluer: So, if I understand correctly that this section is for when we cannot reach the Firestore, so it means that the app is in offline mode. To indicate in the log maybe you can implement an indicator like 

```suggestion
      // Timeout means offline; Firestore will sync later
      Log.i(TAG, "Item is queued for offline mode") 
```

What do you think about it ?

app/src/main/java/com/android/ootd/ui/post/items/AddItemsViewModel.kt
```diff
@@ -263,6 +269,27 @@ open class AddItemsViewModel(
                 _addOnSuccess.value = false
                 return@launch
               }
+
+              // Check if we need to schedule background upload (if offline, we got a local URI)
+              val uri = Uri.parse(uploaded.imageUrl)
```
asunluer: Instead of using `Uri.parse()` you can use this 

```suggestion
              val uri = uploaded.imageUrl.toUri()
```

This is also what AndroidStudio is suggesting when I looked your code.

asunluer (changes requested):
Really good work ! I liked your PR a lot and your implementation. It is clear well explained. 

I have left some suggestions on your implementation. Let me know how they work for you. 

Thank you for spending time on this ! I looked locally also and it seems working. I tried to edit multiple items when I am in offline mode and it uploads the photos correctly. Then, when I connect to internet, it seems no bug also. 

Really great work ! Thanks a lot. 

Additionally, do not forget to fix the warnings that Sonar mentioned if they are not unnecessary changements.

app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
cocoStein: Yes sorry I forgot to remove the debug line in this file !! thanks a lot and yes for the timeout you are right, I've made a `private const val TIMEOUT = 2000L` in the companion obj

app/src/main/java/com/android/ootd/ui/post/items/AddItemsViewModel.kt
cocoStein: That's the problem of coding in vscode, it doesn't show me these tips 😢 

thanks

cocoStein:
Hello thank you for reviewing this complex PR so quickly and It's a good news that it also works flawlessly on your end.

You mentioned to fix the sonar issue, but i don't see any on the sonar report ?

PR #445, opened 2025-12-01
Fix: remove hardcoded pfp and username on posts

bbianca2004:
# Summary
Implemented dynamic user data fetching for the feed to ensure posts always display current usernames and profile pictures instead of outdated cached values from post creation time.

# What

- UI
  - **OutfitPostCard**: Modified to accept `username` and `profilePicUrl` as parameters instead of reading from post object
  - **FeedScreen/FeedList**: Updated to pass user data from `userDataMap` to individual post cards with fallback to cached post data

- ViewModel
  - **FeedViewModel**: 
    - Added `UserRepository` dependency injection
    - Created `UserData` data class to encapsulate username and profile picture
    - Added `userDataMap: Map<String, UserData>` to `FeedUiState` for storing current user information
    - Modified `refreshFeedFromFirestore()` to fetch user data for all post owners from UserRepository
    - Implemented graceful fallback to "Unknown User" when user data is unavailable

- Tests
  - Updated test files to accommodate new parameters and architecture
  - Added test coverage for user data fetching and fallback behavior
# Why

Posts were displaying hardcoded usernames and profile pictures that were saved at post creation time. When users updated their profile information, old posts would continue showing the outdated data, creating inconsistency and confusion. This PR ensures that the changes in user data show up in real time on the feed.

# Screenshots / Videos
https://github.com/user-attachments/assets/932a4d79-5396-4e16-b2fc-6d04440cef66

# Checklist
- [x] Tests: added or updated as appropriate
- [x] Manual verification: 
  - Profile picture updates are reflected in feed immediately
  - Username updates are reflected in feed immediately
  - Graceful handling when user data is missing (shows "Unknown User")

### DIFF
--- a/app/src/test/java/com/android/ootd/ui/feed/FeedViewModelUnitTest.kt
+++ b/app/src/test/java/com/android/ootd/ui/feed/FeedViewModelUnitTest.kt
@@ -5,6 +5,7 @@ import com.android.ootd.model.account.AccountRepository
 import com.android.ootd.model.feed.FeedRepository
 import com.android.ootd.model.posts.LikesRepository
 import com.android.ootd.model.posts.OutfitPost
+import com.android.ootd.model.user.UserRepository
 import com.google.firebase.auth.FirebaseAuth
 import com.google.firebase.firestore.FirebaseFirestore
 import io.mockk.coEvery
@@ -33,6 +34,7 @@ class FeedViewModelUnitTest {
   private lateinit var feedRepository: FeedRepository
   private lateinit var accountRepository: AccountRepository
   private lateinit var likesRepository: LikesRepository
+  private lateinit var userRepository: UserRepository
   private lateinit var firebaseAuth: FirebaseAuth
   private lateinit var firebaseFirestore: FirebaseFirestore
   private val testDispatcher = StandardTestDispatcher()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
asunluer: I was looking to your code and I wondered the definition of `filteredPost` and now I realised that you do not filter the post of your friends that is posted on that day. You filter the current user's post published on that day and all the post of your friends. So, I would suggest you to filter also the posts that your friends posted on that as mentioned in the name. What do you think about it ? 

Additionally, if you filter the posts, published on today, you will not need anymore the `distinct()` function used to define `uniqueOwnerIds` variable.

app/src/main/java/com/android/ootd/ui/feed/FeedScreen.kt
```diff
@@ -172,8 +176,15 @@ fun FeedList(
       val isLiked = likes[post.postUID] ?: false
       val count = likeCounts[post.postUID] ?: 0
 
+      // Get user data for this post owner, fallback to post data if missing
+      val userData = userDataMap[post.ownerId]
+      val username = userData?.username ?: post.name
+      val profilePic = userData?.profilePicUrl ?: post.userProfilePicURL
```
asunluer: What happens when the both condition has blank string ? Shouldn't it better to have a third fallback like 

```suggestion
      val profilePic = userData?.profilePicUrl ?: post.userProfilePicURL ?: ""
```
? 

Let me know what do you think about it

bbianca2004: Good catch, I will do this, thank youuu!

PR #452, opened 2025-12-01
Feat: Public feed v1

cocoStein:
# Summary
This PR adds a public feed were you can see the post of everyone regarding the fact that you are friend or not.
The Post are ordered in decreasing time order. 

# What

- UI
  - Created a new Tab in the feed screen so the user can choose between his `Friend` or `Public` feed.
  - Added a toggle on the `PreviewPost` to choose between posting in private or in public
- ViewModel
  - Added a new `isPublic` field in the Post data class. This is false by default on only if the user chose to it will be added to the public feed.
  - Create a new fun `getPublicFeed` in `FeedRepository.kt` to Retrieves the public feed posts. 
    - I created a `private const val MAX_BATCH_SIZE = 20L` Looking online at example this looks like a good number but it's totally arbitrary.
- Tests
  - Unit tests added/updated according the new implementation of the feed data class and added new connected and unit test to cover all the lines in this PR and all the new functions created in this PR.

# Why
As we tested with user, it would be nice if you could discover new people on a public feed. 

# Screenshots / Videos
This is an example of a complete flow of the new private feed. With a user adding a post on the public feed. As you can see at the moment the public feed also shows the posts that are on your friends list and your own post. We could do user test to see if this is unnecessary. 

https://github.com/user-attachments/assets/fc29f80a-67bf-41df-922c-38b59404b977

# Checklist


app/src/main/java/com/android/ootd/ui/post/OutfitPreviewViewModel.kt
```diff
@@ -155,6 +165,15 @@ class OutfitPreviewViewModel(
...
           val itemIds = items.map { it.itemUuid }
 
+          // If post is public, update items to be public
+          if (state.isPublic) {
+            items.forEach { item ->
+              if (!item.isPublic) {
+                itemsRepository.editItem(item.itemUuid, item.copy(isPublic = true))
+              }
+            }
+          }
```
MarcK0909: Maybe we could do the following to replace the ForEach block you currently have :

Suggested replacement:
val toMakePublic = items.filter { !it.isPublic }

try {
  coroutineScope {
    toMakePublic
      .map { item ->
        async { itemsRepository.editItem(item.itemUuid, item.copy(isPublic = true)) }
      }
      .awaitAll()
  }
} catch (e: Exception) {
  Log.e("OutfitPreviewViewModel", "Failed to mark items public", e)
  // TODO: surface error to UI (update _uiState / show snackbar)
}

Why:
- Only updates non-public items to avoid unecessary operations.
- Runs updates in parallel to reduce total latency.
- Adds error logging and a place to surface failures for easier debugging.

app/src/main/java/com/android/ootd/model/posts/OutfitPost.kt
```diff
@@ -28,6 +29,7 @@ data class OutfitPost(
     val itemsID: List<String> = emptyList(),
     val timestamp: Long = 0L,
     val location: Location = emptyLocation,
+    @get:PropertyName("isPublic") val isPublic: Boolean = false
```
MarcK0909: Maybe add a small comment above this to explain what it is exactly as I had to look up what exactly this code did. I don't think i have encountered this before and it might also be helpful for the others :)

app/src/main/java/com/android/ootd/ui/feed/FeedScreen.kt
```diff
@@ -93,14 +97,48 @@ private fun FeedScaffold(
...
+              rightComposable = {
+                NotificationButton(
+                    onNotificationIconClick,
+                    Modifier.testTag(NAVIGATE_TO_NOTIFICATIONS_SCREEN),
+                    size = 64.dp)
+              })
+          PrimaryTabRow(
+              selectedTabIndex = if (isPublicFeed) 1 else 0,
+              containerColor = MaterialTheme.colorScheme.background,
+              contentColor = MaterialTheme.colorScheme.primary,
```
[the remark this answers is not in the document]
cocoStein: Good catch, fix them !

app/src/main/java/com/android/ootd/ui/post/OutfitPreviewViewModel.kt
cocoStein: Very good idea. thanks for the proposed fix !

cocoStein:
Thanks for the review, I think I will leave it like this and ask some people how they feel about the toggle or if they prefer to have it by default !

PR #457, opened 2025-12-02
Enhancement: Add offline functionality to Feed

asunluer:
# Summary

This PR introduces the implementation of **offline** support to Feed and SeeFit flow. 
Users can now see their friends's posts and view all items associated with a post even when the device is offline, using cached Firestore data and offline-first ViewModel logic. 
This PR is inspired by the work of @cocoStein on his following PRs:  #360,   #439. 

>[!WARNING]
> If post was never cached before, you cannot see the post correctly and when you go to for the first time to SeeFit screen, it cannot neither load the items in offline mode. This is unfortunately unavoidable because Firestore cannot serve data that is never cached. So because of that UI, will show a friendly error message instead of crashing. 
# What

- UI
  - No UI layout changes
  - Feed screen now displays previously-loaded posts while offline.
  - Added a condition on blurred for the public feed. Now it is like this `isBlurred = !hasPostedToday && !isPublicFeed`
  - SeeFit screen now displays associated items while offline if they were previously cached.
- ViewModel
  - FeedViewModel 
     - Implemented **offline-first loading** to the `refreshFeedFromFirestore()` function: 
    - Added two caches for public and friend feed `cachePublicFeed` and `cachePrivateFeed` so that when there is no internet connection it can load from the post there 
    - Added a new function `loadCachedFeed()` which loads cached feed posts from the local cache based on friends and public.
 - SeeFitViewModel 
    - Updated `getItemsForPost(postUuid: String)` function 
- FeedFirestoreRepository
  - Two functions that retrieves the posts from friend and public feed are added: 
    - `getCachedFriendFeed(uids: List<String>)` 
    - `getCachedPublicFeed()`
 - `getFeedForUids(uids: List<String>)` and  `getRecentFeedForUids(uids: List<String>)` are modified. Remove the timeout bound because : 
   -  It was breaking resuming when returning to the screen in offline mode  
   - Causes empty results even when cached data exists
   - Freeze or cancel Firestore queries incorrectly.
- Documentation
  - Kdoc comments updated based on offline logic in: 
    - `refreshFeedFromFirestore()` to **FeedViewModel**
    - `getItemsForPost(postUuid: String)`to **SeeFitViewModel**
- Tests
  - `SeeFitViewModel` unit test is updated
  -  In `FeedRepositoryFirestoreTest` , `getFeed_withCorruptedOrInvalidData_returnsEmptyList()` is updated and renamed to `getFeed_filtersCorruptedOrInvalidData()` 
  - Added new test file `FeedViewModelOfflineTest` to check Feed when it is in offline mode

# Why

Previously, both **Feed** and **SeeFit** required an active internet connection to load data, it was showing too much loading when the user had no internet connection which reduces the UX. Now, with offline mode implementation, user can see their feed in offline mode when there is no connection without seeing the loading screen. 

# Screenshots / Videos

https://github.com/user-attachments/assets/a4dd64f0-9344-4d26-8d69-3983ff144cbb

# Checklist
- [x] Documentation: KDoc added to: 
     - [x] `refreshFeedFromFirestore()` to **FeedViewModel**
     - [x] `getItemsForPost(postUuid: String)`to **SeeFitViewModel**
- [x] Manual verification:
     - [x] Open Feed online -> posts load. 
     - [x]  Change the screen besides Feed screen 
     - [x] Turn off the connection 
     - [x] Go to Feed screen and the posts should be visible 
     - [x] Check a SeeFit of one of your friend -> see an error message 
     - [x] Open the internet connection, reopen again the seeFit screen 
     - [x] Turn off again and. go to seeFit screen this time and items should be visible in offline mode. 

>[!Note]
> This implementation can be improved more. As future work, we can try to implement without using the caches that I defined for this PR

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
+++ b/app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
@@ -94,6 +94,16 @@ open class FeedViewModel(
   /** Refreshes the feed posts from Firestore for the current account. */
   fun refreshFeedFromFirestore() {
     val account = _uiState.value.currentAccount ?: return
+
+    val todayStart = LocalDate.now().atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli()
+
+    val postedOffline =
+        (cachedPublicFeed + cachedPrivateFeed).any { post ->
+          post.ownerId == account.uid && post.timestamp >= todayStart
+        }
+
+    _uiState.value = _uiState.value.copy(hasPostedToday = postedOffline)
+
     viewModelScope.launch {
 
       // 1) Prefill from Firestore local cache immediately
@@ -153,9 +163,6 @@ open class FeedViewModel(
         return@launch
       }
 
-      val todayStart =
-          LocalDate.now().atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli()
-
       val filteredPost =
           if (_uiState.value.isPublicFeed) {
             allPosts
@@ -183,7 +190,7 @@ open class FeedViewModel(
       }
 
       if (_uiState.value.isPublicFeed) {
-        cachedPublicFeed = allPosts
+        cachedPublicFeed = filteredPost
       } else {
         cachedPrivateFeed = filteredPost
       }
@@ -191,10 +198,7 @@ open class FeedViewModel(
       val hasPostedToday =
           withTimeoutOrNull(NETWORK_TIMEOUT_MILLIS) { repository.hasPostedToday(account.uid) }
               ?: false
-      val postedOffline =
-          (cachedPublicFeed + cachedPrivateFeed).any { post ->
-            post.ownerId == account.uid && post.timestamp >= todayStart
-          }
+
       val finalHasPostedToday = hasPostedToday || postedOffline
 
       _uiState.value =


app/src/main/java/com/android/ootd/ui/feed/FeedScreen.kt
```diff
@@ -145,7 +147,7 @@ private fun FeedScaffold(
                     .background(Background)) {
               // Renders the list of posts when user has posted.
               FeedList(
-                  isBlurred = !hasPostedToday,
+                  isBlurred = !hasPostedToday && !isPublicFeed,
```
cocoStein: I see that you changed it so that you can see the public feed even if you haven't posted. 
When implementing the public feed, I didn't think of this, and I like the idea that you can always see the public feed, but then this creates a small issue. If a friend of yours posts on public, then at the moment you can see his post even though you haven’t posted yet. 

Either we make it so that you don't see your friends (and your own) posts on the public, or we change this so that you have to post also to be able to access the public feed.

Let me know what you think, but for me, this works for the moment, and we'll discuss this with the team at a meeting.

app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
```diff
@@ -84,60 +91,113 @@ open class FeedViewModel(
...
+
+      if (cached.isNotEmpty()) {
+        val todayStart =
+            LocalDate.now().atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli()
+
+        val filteredCached =
+            if (_uiState.value.isPublicFeed) cached
+            else
+                cached.filter { post ->
+                  if (post.ownerId == account.uid) post.timestamp >= todayStart else true
```
cocoStein: Great job thinking about the case that in the cache we need to check if the post is expired 🔥

app/src/main/java/com/android/ootd/model/feed/FeedRepositoryFirestore.kt
```diff
@@ -181,6 +169,47 @@ class FeedRepositoryFirestore(private val db: FirebaseFirestore) : FeedRepositor
...
+              .whereEqualTo("isPublic", true)
+              .get(Source.CACHE) // Fetch from cache
+              .await()
+
+      val now = System.currentTimeMillis()
+      val twentyFourHoursAgo = now - MILLIS_IN_24_HOURS
+
+      snap.documents
+          .mapNotNull { mapToPost(it) }
+          .filter { it.timestamp >= twentyFourHoursAgo }
```
cocoStein: It's a small nitpick but why here you create a val for `twentyFourHoursAgo` and the function above you just hard code it ? 

Maybe you could add these values to the companion object that's already existing so we can reuse them in the file !

line 185-186
```kotlin
      results
          .filter { it.timestamp >= System.currentTimeMillis() - MILLIS_IN_24_HOURS }
          .sortedByDescending { it.timestamp }
```

app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
```diff
@@ -84,60 +91,113 @@ open class FeedViewModel(
...
-            Log.e("FeedViewModel", "Failed to load likes for post $postId", e)
+      val filteredPost =
+          if (_uiState.value.isPublicFeed) {
+            allPosts
+          } else {
+            allPosts.filter { post ->
+              if (post.ownerId == account.uid) {
+                post.timestamp >= todayStart
+              } else true
+            }
```
cocoStein: Here I don't understand why you apply this filtering only to the private feed

app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
```diff
@@ -84,60 +91,113 @@ open class FeedViewModel(
...
-                likes = likesMap,
-                likeCounts = countsMap)
-      } catch (e: Exception) {
-        Log.e("FeedViewModel", "Failed to refresh feed", e)
-        _uiState.value =
-            _uiState.value.copy(isLoading = false, errorMessage = "Failed to load feed")
+      if (_uiState.value.isPublicFeed) {
+        cachedPublicFeed = allPosts
+      } else {
+        cachedPrivateFeed = filteredPost
```
cocoStein: Here I don't understand the purpose of this `if else`. Because they both have different cache I don't understand why you need to only update them if you are on the feed.  And this `if else` block is repeated many times across this file.

And on line 155 you make the
```kotlin
      val filteredPost =
          if (_uiState.value.isPublicFeed) {
            allPosts
          } else {
            allPosts.filter { post ->
              if (post.ownerId == account.uid) {
                post.timestamp >= todayStart
              } else true
            }
          }
```

And if I understand correctly then this already checks for the `_uiState.value.isPublicFeed` right ?

So the `cachedPublicFeed = allPosts` could be, with you current setup: `cachedPublicFeed = filteredPost`. Because above you already filter it ?

app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
```diff
@@ -84,60 +91,113 @@ open class FeedViewModel(
...
       }
+
+      val hasPostedToday = repository.hasPostedToday(account.uid)
+      val postedOffline =
```
cocoStein: The posted offline val will be for a later PR when we can add new post offline ?

Because I tried to add a post while being offline and I got stuck on the posting page.
Let me know, if I am doing something wrong !

app/src/main/java/com/android/ootd/ui/feed/FeedScreen.kt
asunluer: Wait a second ! Public feed is always accessible even we haven't posted yet ? 

Because I did this small change from the idea that we do not access on the feed if we have not posted yet. The problem that I encountered while I was implementing if the user was in offline mode, and had already posted, the public feed posts were blurred. Because of that I had this logic.

app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
```diff
@@ -84,60 +91,113 @@ open class FeedViewModel(
...
+          cachedPrivateFeed = filteredCached
+        }
+
+        // Update UI instantly from cache and suppress loading overlay
+        _uiState.value =
+            _uiState.value.copy(
+                feedPosts = filteredCached,
+                hasPostedToday =
+                    (cachedPublicFeed + cachedPrivateFeed).any {
+                      it.ownerId == account.uid && it.timestamp >= todayStart
```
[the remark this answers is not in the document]
asunluer: Yes unfortnately and because of that I separated the posts also and what copilot says is not correct because if the user posted privately, this means that they can see the public feed also but if we toggle to public feed `hasPostedToday` will be false hence the posts will be blurred.

app/src/main/java/com/android/ootd/ui/feed/FeedViewModel.kt
asunluer: Yes you are totally right on this.

app/src/main/java/com/android/ootd/ui/feed/FeedScreen.kt
cocoStein: Ok as you wish, but I still think we can discuss if we want it to be accessible at all time with the other teammate !
That's a good idea

cocoStein (approved):
Thank you for making the changes ! 

And yes don't worry about the offline posting. I was just surprised to see a val in your PR

PR #458, opened 2025-12-02
Fix: Color used and UI element

cocoStein:
# Summary
Minor changes of some colors that were wrong, see an example in the screens bellow. Also changed some buttons to be more stylish and removed the back arrow for the map as it's a top level screen.

# What

- UI
  - Changed the color that were previously `Onbackground` to be hardcoded colors that we have defined, so we don't end up with color we don't know. Because the `Onbackground` changes regarding the phone used
  - Removed the back arrow in the map as it's a top level screen and you have the bottom navigation bar
  - Changed the button in the feed to be a `ExtendedFloatingButton`. More stylish and follow the style of the other floating button in the inventory screen
  - Removed the `Crop` button from the camera screen as it's not yet implemented
- Tests
  - Modified the test according the new changes
# Why

# Screenshots / Videos
 Before:

Screenshot_20251202_202023

Screenshot_20251202_202035

After:
Screenshot_20251202_201904
Screenshot_20251202_201914
Screenshot_20251202_201950

# Checklist
- [ ] Tests: added or updated (unit/instrumented/automation) as appropriate
- [ ] Documentation: README/usage/docs updated where relevant
- [ ] Manual verification: app run and basic flows checked (list steps in "What")

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/map/Map.kt
+++ b/app/src/main/java/com/android/ootd/ui/map/Map.kt
@@ -15,9 +15,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.lifecycle.viewmodel.compose.viewModel
-import com.android.ootd.ui.map.MapScreenTestTags.getTestTagForPostMarker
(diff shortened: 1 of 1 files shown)


MarcK0909 (approved):
looks good to me

PR #469, opened 2025-12-04
Feat: Delete account button

Clemensito:
# Summary
Implemented the UI for the user to be able to delete his account. Upon clicking on the top right icon the user is shown a "Delete account" button, upon clicking on it a confirmation message is shown and if accepted the associated ViewModel deletes the account (by calling the account repository delete method, same for user repository and resets the accounts ViewModel) 
This feature is still in progression because there is no way of safely deleting the user from his friends friend list. This breaks pretty much the app for the user

# What

- UI
  - New pop up on the top right corner of editAccount that shows him the deleteAccount button
 
- ViewModel
  - Created a method that deletes the account. In the Account repo, user repo and signing the user out 
 
- Documentation
  - Kdoc added for delete account method
  - Kdoc added for the new composables (alert dialogs) that have been added
- Tests
  - Integration tests added for both UI & ViewModel

# Why
This allows the user to safely delete his account while still being able to create a new account with the same email address. It ensures no data is saved

# Screenshots / Videos
image

image

# Checklist

### DIFF
--- a/app/src/androidTest/java/com/android/ootd/ui/account/EditAccountScreenTest.kt
+++ b/app/src/androidTest/java/com/android/ootd/ui/account/EditAccountScreenTest.kt
@@ -160,6 +160,8 @@ class EditAccountScreenTest {
     // selectTestTag(UiTestTags.TAG_ACCOUNT_BACK).assertIsDisplayed()
     selectTestTag(UiTestTags.TAG_ACCOUNT_TITLE).assertIsDisplayed()
     selectTestTag(UiTestTags.TAG_ACCOUNT_AVATAR_CONTAINER).assertIsDisplayed()
+    selectTestTag(UiTestTags.TAG_MORE_BUTTON).assertIsDisplayed()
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/ootd/ui/account/EditAccountScreen.kt
```diff
@@ -179,6 +195,14 @@ private fun AccountScreenContent(
             leftComposable = {
               BackArrow(
                   onBackClick = onBack, modifier = Modifier.testTag(UiTestTags.TAG_ACCOUNT_BACK))
+            },
+            rightComposable = {
+              ActionIconButton(
+                  onClick = { showMoreAccountDialog = true },
+                  icon = Icons.Filled.MoreHoriz,
+                  contentDescription = "More options",
+                  tint = colorScheme.onBackground,
```
cocoStein: As we never properly set up the logic for the `onBackground` for the light / dark mode, I think it would be better if you just hardcode one of the predefined colors we have in our color scheme so that on a real device it looks as we expect it to look.

PR #492, opened 2025-12-08
Feat: Display public profiles on the map 

j-meric:
# Summary
Implement feature to display public user profiles on the map with clustered markers and generic rendering infrastructure to avoid duplication. This is not the final version yet, and proper navigation from the public profiles will be added as well as a distinction between public profiles that are friends and those that are not depending on results of UX tests. 

# What
- UI
  - New generic `ClusteredMap<T>` composable centralizing clustering, camera handling, and click behavior
  - Refactored `FriendsPostsMap` and `FindFriendsMap` as thin configuration layers
  - Introduced `ProfileMarkerItem` interface; updated `PostMarker` and `PublicProfileMarker` implementations
  - Fixed map markers not showing by collecting UI state and passing data to ViewModel APIs
  - Removed `rememberCoroutineScope()` from `Map.kt` to avoid lifecycle leaks
- ViewModel
  - Centralized location adjustment logic with generic `withAdjustedLocations()` helper in `MapViewModel`
  - Added `Locatable` and `ItemWithAdjustedLocation` abstractions
- Data Model
  - Added `PublicLocation` data class and public location APIs to `AccountRepository`
  - Implemented public location syncing in both in-memory and Firestore implementations
  - Added Firestore rules to expose `publicLocations` collection with read access
  - Added `InvalidLocationException` for validation
- Rendering
  - Extracted generic `ProfileClusterRenderer<T>` base class
  - Refactored post and public profile renderers to extend it, centralizing image loading and caching
  - Renderers now own internal `CoroutineScope` instead of accepting external scopes
- Tests
  - Added unit tests for `PublicProfileMarker`, `PostMarker`, and public location filtering in `MapViewModel`
  - Added Firestore and in-memory repository tests for public location syncing
  - Simplified test documentation and assertions

# Screenshot
image

# Why
Addresses feature request to display public user profiles on the map while maintaining privacy constraints. The generic rendering framework eliminates duplication and makes adding future locatable item types straightforward. The map marker fix ensures UI recomposition occurs when async data loads. 

# Disclaimer
Higher Sonar coverage is unfortunately not possible, as methods in clusterRenderer are not covered as the Google Maps SDK's clustering internals may invoke its methods through native code. Ie. methods like `onBeforeClusterItemRendered()`, `onClusterItemUpdated()`, and `onBeforeClusterRendered()` are called asynchronously by the clustering library's internal rendering loop, not directly by test code, which makes it impossible for Sonar to detect.

# Checklist
- [x] Documentation: KDoc comments updated where relevant
- [x] Manual verification: app run and basic flows checked (public location visibility, privacy toggle behavior, map clustering)

### DIFF
--- a/app/src/main/java/com/android/ootd/model/account/AccountRepositoryInMemory.kt
+++ b/app/src/main/java/com/android/ootd/model/account/AccountRepositoryInMemory.kt
@@ -21,48 +21,42 @@ class AccountRepositoryInMemory : AccountRepository {
                   ownerId = "user1",
                   username = nameList[0],
                   profilePicture = "u1.jpg",
-                  friendUids = listOf("user2", "user3"),
-                  isPrivate = true),
+                  friendUids = listOf("user2", "user3")),
           "user2" to
               Account(
                   uid = "user2",
                   ownerId = "user2",
                   username = nameList[1],
                   profilePicture = "u2.jpg",
-                  friendUids = listOf("user1"),
-                  isPrivate = true),
+                  friendUids = listOf("user1")),
           "user3" to
               Account(
                   uid = "user3",
                   ownerId = "user3",
                   username = nameList[2],
                   profilePicture = "u3.jpg",
-                  friendUids = emptyList(),
-                  isPrivate = true),
+                  friendUids = emptyList()),
           "user4" to
               Account(
                   uid = "user4",
                   ownerId = "user4",
                   username = nameList[3],
                   profilePicture = "u4.jpg",
-                  friendUids = listOf("user1", "user2"),
-                  isPrivate = true),
+                  friendUids = listOf("user1", "user2")),
           "user5" to
               Account(
                   uid = "user5",
                   ownerId = "user5",
                   username = nameList[4],
                   profilePicture = "u5.jpg",
-                  friendUids = emptyList(),
-                  isPrivate = true),
+                  friendUids = emptyList()),
           "nonRegisterUser" to
               Account(
                   uid = "nonRegisterUser",
                   ownerId = "nonRegisterUser",
                   username = "",
                   profilePicture = "u0.jpg",
-                  friendUids = listOf(),
-                  isPrivate = true))
+                  friendUids = listOf()))
 
   private val accountUpdates = MutableStateFlow<Pair<String, Account?>>(Pair("", null))
   private val publicLocations = mutableMapOf<String, PublicLocation>()


app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
```diff
@@ -321,40 +342,54 @@ class AccountRepositoryFirestore(
...
-      db.collection(ACCOUNT_COLLECTION_PATH).document(userID).delete().await()
-      Log.d(TAG, "Successfully deleted account with UID: $userID")
-    } catch (_: NoSuchElementException) {
-      throw UnknowUserID()
-    } catch (e: Exception) {
-      Log.e(TAG, "Error deleting account: ${e.message}", e)
-      throw e
+    try {
+      db.collection(PUBLIC_LOCATIONS_COLLECTION_PATH).document(userID).delete().await()
+    } catch (_: Exception) {
```
stefantaga24: I think here it is better to first check if it exists or not. If the internet connection of the user goes out or something like that, I think the error should appear, here we ignore it completely.

app/src/main/java/com/android/ootd/model/account/AccountRepositoryInMemory.kt
```diff
@@ -21,53 +21,59 @@ class AccountRepositoryInMemory : AccountRepository {
                   ownerId = "user1",
                   username = nameList[0],
                   profilePicture = "u1.jpg",
-                  friendUids = listOf("user2", "user3")),
+                  friendUids = listOf("user2", "user3"),
+                  isPrivate = true),
```
stefantaga24: Considering isPrivate is now by default true, I think you can remove the added lines in this file for making isPrivate true everywhere.

app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
j-meric: Good point, I reworked the logic according to your suggestion, ie. I am checking first if the location exists first.

app/src/main/java/com/android/ootd/model/account/AccountRepositoryInMemory.kt
j-meric: Great catch! I corrected the code. To be honest I changed the the default to isPrivate after having done the modifications in AccountRepositoryInMemory and then forgot to change it back.

stefantaga24 (approved):
Thank you for implementing the changes, LGTM!

PR #507, opened 2025-12-10
Refactor user search screen UI

stefantaga24:
# Summary

Make search screen simpler and move the following mechanism on the view user screen.

# What

- UI
  - Removed the dropdown functionality and now the list appears every time a user searches
  - The follow button on user view respects the "Follow" -> "Requested" -> "Unfollow" transformation.

- ViewModel
  - The UserSearchScreen view model was simplified.
  - The follow button is moved to the view user view model.
- Tests
  - Updated tests related to user search screen and removed extra tests
  - The second end to end test cannot reliably test following mechanism so I removed it. This is 
ok as it is tested both in third and fourth end to end test.

# Why

The previous design was very clunky and the profile card was taking space without any function as the
user view is much better at displaying information.

# Screenshots / Videos

To manually test , search for a user and try to follow them:

https://github.com/user-attachments/assets/fa8056d7-1064-4ef8-a3a5-9adf385c7cd2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: "temurin"
-          java-version: "17"
+          java-version: "21"
 
       - name: Gradle cache
         uses: gradle/actions/setup-gradle@v3
@@ -67,7 +67,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: "temurin"
-          java-version: "17"
+          java-version: "21"
 
       - name: Gradle cache
         uses: gradle/actions/setup-gradle@v3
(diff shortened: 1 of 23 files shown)


Clemensito (commented):
This PR looks really good, you did a fantastic job implementing these changes. There aer some comments I'd like you to address as well as the sonar issue. But other than that LGTM 🔥

app/src/main/java/com/android/ootd/ui/search/UserSelectionField.kt
```diff
@@ -1,100 +1,132 @@
...
+fun UserSuggestionItem(user: User, onClick: () -> Unit, modifier: Modifier = Modifier) {
+  Row(
+      modifier =
+          modifier
+              .fillMaxWidth()
+              .clickable(onClick = onClick)
+              .padding(horizontal = 16.dp, vertical = 12.dp),
+      verticalAlignment = Alignment.CenterVertically) {
+        // Profile image
+        ProfilePicture(size = 48.dp, profilePicture = user.profilePicture, username = user.username)
```
Clemensito: Here I would suggest you reduce the font size of the font inside the profile picture. It is way too big and does not fit inside the circle

Clemensito (approved):
Thanks for your quick fixes and improvements ! LGTM 🔥

PR #517, opened 2025-12-11
Enhancement : Image are displaying less cropped

asunluer:
# Summary

This PR is followed by the previous PR #505. Now the images are less cropped and the portrait photos are uploaded more properly in the feed. 
This one also fixes the error where loading the images in the Feed takes too much time. 
# What

- OutfitPostCard
  - Add an aspectRatio instead of a fixed height. 
  - Change the `.fillMaxSize` to `.fillMaxWidth`
  - Change the `ContentScale.Fit` to `ContentScale.Crop`
- Tests
      - No additional tests are added since this changes how the images are displayed
      - **TestingUtils** : In the `checkItemAppearsInPost` test, it checks the existing and visibility of **See Fit** button. To make it visible this button, I activate the scrollability of the screen and made true the `useUnmergedTree` parameter for `clickWithWait` function. 

# Why

Before the images were too cropped, which was cutting too much some part of the images and since the users are mostly uploading their photos in portrait version, this version fits better. 
The images were not loaded immediately in the **Feed** because of the **AsyncImage** composable because this composable detects the image sizes at run time meaning that if we do not have a constant image size, it will try to figure it out buy itself which takes too much time. So to avoid this situation, I used a fixed aspect ratio value, so that `AsyncImage` can know what will be the image size and based on that it can resize the images quickly which avoids to see overlapped cards in the **Feed*. 

# Screenshots / Videos

Before, it was like this 

Screenshot 2025-12-08 at 09 47 33

Now it is like that 

Screenshot 2025-12-12 at 00 17 13

# Checklist
-[X] : Manual verification : Go to feed and see how the images are displayed.

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/feed/OutfitPostCard.kt
+++ b/app/src/main/java/com/android/ootd/ui/feed/OutfitPostCard.kt
@@ -174,9 +174,10 @@ private fun PostImage(post: OutfitPost, isBlurred: Boolean, modifier: Modifier =
           Modifier.fillMaxWidth()
               .clip(RoundedCornerShape(12.dp))
               .background(White)
-              .height(500.dp)
+              .aspectRatio(3f / 4f)
               .testTag(OutfitPostCardTestTags.POST_IMAGE_BOX)
-              .then(modifier)) {
+              .then(modifier),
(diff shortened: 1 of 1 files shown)


MarcK0909 (approved):
LGTM

PR #532, opened 2025-12-14
Add find friends on map button in user search screen

stefantaga24:
# Summary

Added a way to go from the user search screen to the find friends part of the map.

# What

- UI
  - Added small text on the user search screen in order to go to finding friends on the map.
  - Changed the back arrow to be the same with the one from our theme.

# Why

In this way, the user can also use the map functionality.

# Screenshots / Videos

The text added can be seen in this screenshot:
image

### DIFF
--- a/app/src/main/java/com/android/ootd/ui/search/UserSelectionField.kt
+++ b/app/src/main/java/com/android/ootd/ui/search/UserSelectionField.kt
@@ -11,6 +11,7 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.unit.dp
 import com.android.ootd.model.user.User
 import com.android.ootd.ui.theme.Background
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/ootd/ui/search/UserSelectionField.kt
```diff
@@ -43,14 +44,9 @@ fun UserSelectionField(
     Row(
         modifier = Modifier.fillMaxWidth().padding(16.dp),
         verticalAlignment = Alignment.CenterVertically) {
-          IconButton(
```
j-meric: **Praise**: Well done using our theme's back arrow for consistency!

app/src/main/java/com/android/ootd/ui/search/UserSelectionField.kt
```diff
@@ -84,6 +80,15 @@ fun UserSelectionField(
               singleLine = true)
         }
 
+    Text(
+        text = "Click to find friends on the map",
```
j-meric: **Question**: What do you think of writing "Click to find public friends on the map" for more clarity, so that users know that not all the friends are displayed on the map?

j-meric (changes requested):
Nice addition to the app, this will definitely enhance user flow. . However, one test in UserSearchScreenTest is still failing. Furthermore, I have got two questions with regards to the clickable text. Once fixed, the code is good to go.

app/src/main/java/com/android/ootd/ui/search/UserSelectionField.kt
stefantaga24: Thank you!

j-meric (approved):
Thanks for the quick changes.  The tests seem to pass, but the CI seemed to have hung up, so could you confirm that the CI passes before merging?

PR #539, opened 2025-12-14
 Add friendCount to the view user page

stefantaga24:
# Summary

Add friendCount in the user repository as it is public to every user

# What

- ViewModel
  - In the viewUser page, the friendCount is taken from the user repository.
  - The friendCount is updated in the addFriend and removeFriend functions in the account repository.
- FirebaseRules
  - Created new rule for the user repository such that if another user wants to add a friend (by accepting the notification), they can modify the friendCount of the other user.
  - Created new rule for the user repository such that if another user removes a friend, they are able to change the friendCount of the other user.
- Tests
  - Added a check in the fourth end to end test to make sure that the friendCount is updated for both users when the friendship is accepted.

# Why

The friendCount was implemented in the viewUser page but was not functional before.

# Screenshots / Videos

### DIFF
--- a/app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
@@ -262,20 +262,28 @@ class AccountRepositoryFirestore(
         throw NoSuchElementException("The user with id $friendID not found")
       }
 
-      val userRef = db.collection(ACCOUNT_COLLECTION_PATH).document(userID)
+      val currentUserAccountRef = db.collection(ACCOUNT_COLLECTION_PATH).document(userID)
 
       // With arrayUnion there can be no duplicates
       // https://firebase.google.com/docs/firestore/manage-data/add-data , Update elements in an
       // array section
 
       // Add the friendID in your friend list
-      userRef.update("friendUids", FieldValue.arrayUnion(friendID)).await()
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/ootd/model/account/AccountRepositoryFirestore.kt
```diff
@@ -262,20 +262,28 @@ class AccountRepositoryFirestore(
...
       // Add yourself to the friendID's friend list.
       // This works because we have not deleted the follow notification yet.
-      val friendRef = db.collection(ACCOUNT_COLLECTION_PATH).document(friendID)
+      val friendAccountRef = db.collection(ACCOUNT_COLLECTION_PATH).document(friendID)
+      val friendUserRef = db.collection(USER_COLLECTION_PATH).document(friendID)
+
       try {
-        friendRef.update("friendUids", FieldValue.arrayUnion(userID)).await()
+        friendAccountRef.update("friendUids", FieldValue.arrayUnion(userID)).await()
+        friendUserRef.update("friendCount", FieldValue.increment(1)).await()
```
bbianca2004: Question: These updates to friendUids (account collection) and friendCount (user collection) are done in separate calls.
What happens if one of the two operations fails(unlikely but still), from what I see the other will still go through right?

bbianca2004: Okay, I fully understand now, thank you so much for your very detailed explanations and for making the time to answer my questions. Everything looks great to me, you can safely go ahead and merge it :))

PR #571, opened 2025-12-17
fix(items): Fix public items not loading

stefantaga24:
# Summary

There was a problem with public posts because we could not fetch the items to it.

# What

- ViewModel
  - We now check if the post is public, we need to retrieve only public items.
- Firebase rules
  - We now also check for public == true in our firebase rules.  

# Why

This was hurting the UX experience.


cocoStein (approved):
thanks for the fix !!!

